PR #2, opened 2019-10-30
HT-1618: add docker container for running validation

aelkiss:
I'm not a docker expert, so I would like some feedback on this.


daaang (approved):
Your dockerfiles are clear, idempotent, and they don't have side effects that I can see. They also don't appear to depend needlessly on any particular build environment. I guess they don't look especially configurable, but configurability seems to me like something to be added when needed, as needed. LGTM

PR #5, opened 2019-11-19
PFDR-219: add support for validating WAVE files with jhove 1.20; drop support for jhove 1.6

aelkiss:


botimer (approved):
This looks good, though I am a little unclear about the `git` tag for the images. Is that analogous to `latest`?

aelkiss:
Yes, perhaps it should just be 'latest' instead. I'm not sure what the standard/best practices for tags are.

aelkiss:
I've changed the tag in the wrapper scripts and the subdirectory name for the dockerfile that builds an image with master from github.

PR #6, opened 2020-02-19
HT-2251: remove remediation for samplesperpixel, bitspersample

aelkiss:
This was leading to cases where images with 4 bits per pixel were
automatically getting downsampled to 1 bit per pixel, when we would
rather just reject them.

This will now:

- only convert TIFF to JPEG2000 if BitsPerSample = 8 and SamplesPerPixel
= 3 or SamplesPerPixel = 1

- only remediate other problems in TIFFs with ImageMagick if the image
is bitonal (BitsPerSample = 1 and SamplesPerPixel = 1)

- not attempt to remediate other problems with SamplesPerPixel

### DIFF
--- a/lib/HTFeed/Stage/ImageRemediate.pm
+++ b/lib/HTFeed/Stage/ImageRemediate.pm
@@ -227,9 +227,7 @@ sub _remediate_tiff {
                                               # wrong data type for tag - will get automatically stripped
                                                'Type mismatch for tag',
                                                # related to thumbnails, which imagemagick will strip
-                                               'JPEGProc not defined for JPEG compression',
-                                               'For PhotometricInterpretation, SamplesPerPixel must be >= 3, equals1',
-                                               'For PhotometricInterpretation, SamplesPerPixel must be >= 1, equals: 1');
+                                               'JPEGProc not defined for JPEG compression');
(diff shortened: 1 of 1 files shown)


lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -258,7 +256,8 @@ sub _remediate_tiff {
 
     # Does it look like a contone? Bail & convert to JPEG2000 if so.
     if(!$bad and $fields->{'IFD0:BitsPerSample'} eq '8'
-            or $fields->{'IFD0:SamplesPerPixel'} eq '3') {
+        and ($fields->{'IFD0:SamplesPerPixel'} eq '3'
+             or $fields->{'IFD0:SamplesPerPixel'} eq '1')) {
```
daaang, line 260: Took me a minute to parse out where the parentheses are, but I think this looks right.

daaang (approved):
Yeah this looks like the simplest change.

lib/HTFeed/Stage/ImageRemediate.pm
aelkiss: yeah we don't have rubocop to fix our terrible formatting. the exiftool fields are definitely ripe for some class extraction as a wrapper around them rather than treating them as a hash, but not today...

PR #9, opened 2020-04-28
HT-1753: fixes for checksum.md5 handling in depositor-submitted material

aelkiss:
In specific adds tests (and bug fixes) for several long-standing issues regarding `checksum.md5`:
* better error message when checksum.md5 is missing
* provokes errors when there are files in the package that are not in checksum.md5

This also involves some work to make it easier to write tests that involve testing log output from running stages on small fixtures:

* Factors out setup to module ( à la `spec_helper.rb`) that can be used by future `Test::Spec` tests
* Support for testing log messages by logging to a string & then checking the results - not ideal in that it would be nicer to intercept the structured logging, but it's a start.


lib/HTFeed/Log/Layout/PrettyPrint.pm
```diff
@@ -28,17 +28,19 @@ sub new {
 }
 
 sub render {
-    my($self, $message, $category, $priority, $caller_level) = @_;
+    my($self, $incoming_message, $category, $priority, $caller_level) = @_;
```
aelkiss, line 31: Changes here are because previously this method mutated the incoming message; now it duplicates it so that mutation doesn't affect the message for other appenders that might be in use

lib/HTFeed/Volume.pm
```diff
@@ -244,7 +245,11 @@ sub get_checksum_md5 {
 
     my $checksums = {};
 
-    open(FILE, $checksum_path) or die $!;		
+    unless ( -e $checksum_path ) {
+      $self->set_error("MissingFile", file => $checksum_file);
+    }
+
+    open(FILE, $checksum_path) or die($!);
```
aelkiss: We could also have changed this to `die("Can't open $checksum_path: $!")`, which still might be better

PR #12, opened 2020-05-20
HT-1687: Copy zip to staging area, validate before move to final location

aelkiss:
This means we should spend less time chasing database problems that are happening only because we're using a different database in development than in production.


bin/wait-for
```diff
@@ -0,0 +1,80 @@
...
+  $cmdname host:port [-t timeout] [-- command args]
+  -q | --quiet                        Do not output any status messages
+  -t TIMEOUT | --timeout=timeout      Timeout in seconds, zero for no timeout
+  -- COMMAND ARGS                     Execute command with args after the test finishes
+USAGE
+  exit "$exitcode"
+}
+
+wait_for() {
+  command="$*"
```
daaang, line 47: `$*` (as opposed to `$@`) always makes me nervous, but given that this is destined for `exec $command`, I think it should be fine.

docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
     environment:
       - HTFEED_CONFIG=/usr/local/feed/etc/config_test.yml
       - FEED_HOME=/usr/local/feed
-    command: make test TEST_VERBOSE=1
+    command: bin/wait-for --timeout=300 mariadb:3306 -- make test TEST_VERBOSE=1
```
daaang, line 19: I can't explain why, but I love this.

daaang (approved):
It looks to me like this is a change purely to the testing environment, so that tests are run against mariadb instead of sqlite3. It also looks like groundwork is laid for testing database behavior around collated items, but nothing yet involving changes to the source code.

If that's right, then this looks good to me!

bin/wait-for
aelkiss: This is just cribbed from https://github.com/eficode/wait-for; it should be cited in the code.

docker-compose.yml
aelkiss: Cribbed from the "docker for rails" book, which recommends using this approach for a one-shot db:migrate container

aelkiss:
I think this can be merged after re-review by @daaang. The next step is to pull out the stage, validate, link, and move operations to a separate class where we can independently test & vary the steps (e.g. for data den, direct scp/rsync to ictc, etc)

aelkiss:
I'll manually test this out end-to-end and make sure that everything works as expected, then will go ahead and merge.

PR #13, opened 2020-07-07
HT-2433: support ingest to multiple storage backends

aelkiss:
In particular, supports ingest both to a local pairtree and a "versioned" pairtree


aelkiss:
I've tested this with Internet Archive and Google volumes, and @daaang  has validated the general approach. There's some follow-on cleanup to do, but this should be a good enough start go ahead and start depositing some material to Data Den.

PR #16, opened 2020-07-29
HT-2518 Move configuration for storage classes

moseshll:
…e class rather than global

CAVEATS

- `Volume.pm` is still using the `repository` version of the `link_dir` config since it doesn't have access to a storage instance. I assume that can be fixed on a related ticket like HT-2519.
- `collate.t` uses `repository` config when creating a set of storage classes on the fly.

### DIFF
--- a/etc/config_ingest_docker.yml
+++ b/etc/config_ingest_docker.yml
@@ -25,11 +25,7 @@ storage_classes:
     # The directory into which volumes will be loaded
     obj_dir: /sdr1/obj
   - class: HTFeed::Storage::VersionedPairtree
-    # The directory in which symbolic links will be created to each volume
-    link_dir: /sdr2/obj
-    # The directory into which volumes will be loaded
-    obj_dir: /sdr1/obj
-    backup_obj_dir: /htdataden
+    obj_dir: /htdataden
 
 jhove: /usr/bin/jhove
 jhoveconf: /etc/jhove/jhove.conf
--- a/etc/config_ingest_docker.yml
+++ b/etc/config_ingest_docker.yml
@@ -1,13 +1,5 @@
 feed_home: /usr/local/feed
 
-# for end-to-end testing in docker
-repository:
-  # The directory in which symbolic links will be created to each volume
-  link_dir: /sdr2/obj
-  # The directory into which volumes will be loaded
-  obj_dir: /sdr1/obj
-  backup_obj_dir: /htdataden
-
 staging_root: /tmp/ingest
 sip_root: /tmp/stage
 
@@ -27,8 +19,17 @@ xerces: /usr/local/feed/blib/bin/validateCache
 epubcheck: /usr/bin/java -jar /usr/bin/epubcheck
 
 storage_classes:
-  - HTFeed::Storage::LocalPairtree
-  - HTFeed::Storage::VersionedPairtree
+  - class: HTFeed::Storage::LocalPairtree
+    # The directory in which symbolic links will be created to each volume
+    link_dir: /sdr2/obj
+    # The directory into which volumes will be loaded
+    obj_dir: /sdr1/obj
+  - class: HTFeed::Storage::VersionedPairtree
+    # The directory in which symbolic links will be created to each volume
+    link_dir: /sdr2/obj
+    # The directory into which volumes will be loaded
+    obj_dir: /sdr1/obj
+    backup_obj_dir: /htdataden
 
 jhove: /usr/bin/jhove
 jhoveconf: /etc/jhove/jhove.conf
--- a/etc/config_test.yml
+++ b/etc/config_test.yml
@@ -32,7 +32,6 @@ storage_classes:
     link_dir: /tmp/obj_link
     # The directory into which volumes will be loaded
     obj_dir: /tmp/obj
-    backup_obj_dir: /tmp/obj_backup
 
 jhove: /usr/bin/jhove
 jhoveconf: /etc/jhove/jhove.conf
--- a/etc/config_test.yml
+++ b/etc/config_test.yml
@@ -27,7 +27,12 @@ repository:
   backup_obj_dir: /tmp/obj_backup
 
 storage_classes:
-  - HTFeed::Storage::LocalPairtree
+  - class: HTFeed::Storage::LocalPairtree
+    # The directory in which symbolic links will be created to each volume
+    link_dir: /tmp/obj_link
+    # The directory into which volumes will be loaded
+    obj_dir: /tmp/obj
+    backup_obj_dir: /tmp/obj_backup
 
 jhove: /usr/bin/jhove
 jhoveconf: /etc/jhove/jhove.conf
--- a/lib/HTFeed/Stage/Collate.pm
+++ b/lib/HTFeed/Stage/Collate.pm
@@ -31,7 +31,8 @@ sub storages_from_config {
 
   my @storages;
   foreach my $storage_class (@{get_config('storage_classes')}) {
-    push(@storages, $storage_class->new(volume => $self->{volume}));
+    push(@storages, $storage_class->{class}->new(volume => $self->{volume},
+                                                 config => $storage_class));
   }
 
   return @storages;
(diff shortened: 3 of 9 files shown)


aelkiss:
Right, `Volume` will have to continue to use the old configuration until we do HT-2521.

etc/config_ingest_docker.yml
```diff
@@ -27,8 +19,17 @@ xerces: /usr/local/feed/blib/bin/validateCache
...
 storage_classes:
-  - HTFeed::Storage::LocalPairtree
-  - HTFeed::Storage::VersionedPairtree
+  - class: HTFeed::Storage::LocalPairtree
+    # The directory in which symbolic links will be created to each volume
+    link_dir: /sdr2/obj
+    # The directory into which volumes will be loaded
+    obj_dir: /sdr1/obj
+  - class: HTFeed::Storage::VersionedPairtree
+    # The directory in which symbolic links will be created to each volume
```
aelkiss: I probably didn't give enough context to what these were; for `VersionedPairtree` the preferred form for the config is:

```yaml
- class: HTFeed::Storage::VersionedPairtree
  obj_dir: /htdataden
```

etc/config_test.yml
```diff
@@ -27,7 +27,12 @@ repository:
   backup_obj_dir: /tmp/obj_backup
 
 storage_classes:
-  - HTFeed::Storage::LocalPairtree
+  - class: HTFeed::Storage::LocalPairtree
+    # The directory in which symbolic links will be created to each volume
+    link_dir: /tmp/obj_link
+    # The directory into which volumes will be loaded
+    obj_dir: /tmp/obj
+    backup_obj_dir: /tmp/obj_backup
```
aelkiss: Don't need `backup_obj_dir` for `LocalPairtree`

lib/HTFeed/Stage/Collate.pm
```diff
@@ -31,7 +31,8 @@ sub storages_from_config {
 
   my @storages;
   foreach my $storage_class (@{get_config('storage_classes')}) {
-    push(@storages, $storage_class->new(volume => $self->{volume}));
+    push(@storages, $storage_class->{class}->new(volume => $self->{volume},
```
aelkiss: I had made a comment about maybe adding a factory method to `HTFeed::Storage`, but ignore it;  I don't think it improves the legibility that much.

aelkiss: I would change `$storage_class` to `$storage_config` though

t/lib/HTFeed/Test/TempDirs.pm
```diff
@@ -59,6 +59,11 @@ sub setup_example {
   foreach my $dirtype ($self->repo_dirtypes) {
     $self->{$dirtype} = tempdir("$tmpdir/feed-test-$dirtype-XXXXXX");
     set_config($self->{$dirtype},'repository',$dirtype);
+    my $storage_classes = get_config('storage_classes');
```
aelkiss: Hmm, this is a bit messy.. I had forgotten about `TempDirs` assuming a global config. I think this is OK for use in `collate.t` (where we have to set the global storage class config anyway) but we could perhaps consider instead making it `collate.t`'s responsibility to call `set_config(...,'storage_classes')`...

t/storage.t
```diff
@@ -40,21 +53,21 @@ describe "HTFeed::Storage" => sub {
       my $volume = stage_volume(@_);
 
       my $storage = HTFeed::Storage::LocalPairtree->new(
-        volume => $volume);
+        volume => $volume,
+        config => config_for_class('HTFeed::Storage::LocalPairtree'));
```
aelkiss: This needs to be untangled a little bit. I *think* we could probably change TempDirs to spit back a config that can be passed to this instead of needing to fuss with the global storage class config in `config_for_class`

t/storage.t
```diff
@@ -40,21 +53,21 @@ describe "HTFeed::Storage" => sub {
...
       return $storage;
     }
 
     sub make_old_version_other_dir {
       my $storage = shift;
-
-      my $real_obj_dir = get_config('repository','obj_dir');
-      my $other_obj_dir = get_config('repository','other_obj_dir');
-      set_config($other_obj_dir,'repository','obj_dir');
+      my $real_obj_dir = $storage->get_storage_config('obj_dir');
```
aelkiss: I think we want to change this to build a separate storage instance with the "other obj dir" (which we could get from our TempDirs object that just knows about the directories rather than fussing with the global config)

t/storage.t
```diff
@@ -159,26 +172,17 @@ describe "HTFeed::Storage" => sub {
...
-          set_config(get_config('repository','link_dir'), 'repository','obj_dir');
-        };
-
-        after each => sub {
-          set_config($old_obj_dir,'repository','obj_dir');
-        };
-
         it "works" => sub {
           my $storage = local_storage($tmpdirs,'test','test');
+          $storage->set_storage_config(config_for_class('HTFeed::Storage::LocalPairtree')->{link_dir},'obj_dir');
```
aelkiss: Again I think we want to construct a new instance of the storage with the desired config up front rather than mutating it afterwards

aelkiss (changes requested):
There's a bunch of confusion and added complexity primarily because of `TempDirs` changing global config and various things especially in `storage.t` mutating global config. I tried to give a few pointers towards cleaning that up but I can also dive in if you get stuck on it.

etc/config_ingest_docker.yml
aelkiss: (There's probably a corresponding change in `VersionedPairtree` required to use that config key instead of `backup_obj_dir`)

moseshll: `VersionedPairtree` uses `'backup_obj_dir'` in two places: `object_path()` and `stage_path()`. Am I correct in suspecting that `'backup_obj_dir'` as a config key can be considered obsolete (aside from possible legacy use in testing)?

t/lib/HTFeed/Test/TempDirs.pm
moseshll: `storage.t` also relies on the `TempDirs.pm` hack, so actually it seems reasonable to me to keep the responsibility for mucking with the global config there.

t/storage.t
moseshll: I was able to do this as long as I made sure to check whether `tempdir` had already been called for a given directory type. Still seems kind of brittle. Here's what I added:
```
sub config_for_storage_class {
  my $self  = shift;
  my $class = shift;

  my $tmpdir = $self->{tmpdir};
  my $config = {};

  foreach my $dirtype ($self->repo_dirtypes) {
    unless (exists $self->{$dirtype}) {
      $self->{$dirtype} = tempdir("$tmpdir/feed-test-$dirtype-XXXXXX");
    }
    $config->{$dirtype} = $self->{$dirtype};
  }
  return $config;
}
```

etc/config_ingest_docker.yml
aelkiss: Correct, any places that use `backup_obj_dir` should be changed to just use `obj_dir`

t/lib/HTFeed/Test/TempDirs.pm
aelkiss: Right, I was hoping that `storage.t` could avoid having to fuss with the global config at all (i.e. have it only rely on storage classes constructed with the particular config they need rather than having to look in the global config)

t/storage.t
aelkiss: That's probably OK, although worth verifying if there are actually cases where the tests rely on having the same temporary directories between two storage instances or not.

t/storage.t
moseshll: I think the two instances are both getting config from `$tmpdirs` now with my second commit.

moseshll:
Commit 2 is as far as I could take things before they started _really_ going off the rails. I don't really have any business trying to do more on this project this week (prolly shoulda taken the whole week off).

aelkiss:
Sounds good, I'll take it from here.

aelkiss (approved):

PR #17, opened 2020-07-29
HT-2520: Validate zip contents before staging

aelkiss:
The original issue specifies validating the zip contents in pack, but on
reflection that won't really work since the METS isn't constructed yet.
We could run the validation at the end of the METS stage or in an
additional stage before handle, but running at the beginning of collate
(i.e. before staging) seemed like a less disruptive change for the time
being.


moseshll (approved):
Looks and runs fine here (as long as my local t/classtests.t hack is left intact).

PR #19, opened 2020-10-30
HT-2697 Download & queue material from AWS S3 bucket for FRAME/EMMA

moseshll:
First pass.
Using stub library t/lib/AWS.pm to return canned data so as not to rely on AWS credentials and binaries.
I don't like the fact that the actual AWS commands are not part of the tests.
Tests concentrate on object list pagination and getting materials downloaded to the right place.
Not sure about how `enqueue` will be invoked.


aelkiss (commented):
I will merge this in with some work I'm doing on HT-2523 for upload to S3 -- in particular see https://github.com/hathitrust/feed/commit/d29ff51cb0f832cb76ce2997559f4e5ec8355ee6#diff-634ab5f06edb4c91437c43f0c2309d92d28a047f9f68c583bba6b1e412d1d8d4 

It's using the same approach for shelling out to `aws s3api` - just need to put everything in the same place

aelkiss:
Started work towards integrating, see https://github.com/hathitrust/feed/tree/HT-2697-frame-download-s3-merged branch. Should be able to finish that up tomorrow.

aelkiss:
Superceded by #26

PR #20, opened 2020-11-06
HT-2523: Support for encrypted zip in backups

aelkiss:
- Encrypts the zip in the staging area
- Validates that it can decrypt it before moving it to staging area
- Records the md5sum of the encrypted zip in the backups table (to allow
verification without decryption)

### DIFF
--- a/docker/database/sql/000_feed_schema.sql
+++ b/docker/database/sql/000_feed_schema.sql
@@ -117,6 +117,7 @@ CREATE TABLE IF NOT EXISTS `feed_backups` (
   `version` varchar(16) DEFAULT NULL,
   `zip_size` bigint(20) DEFAULT NULL,
   `mets_size` bigint(20) DEFAULT NULL,
+  `saved_md5sum` char(32) DEFAULT NULL,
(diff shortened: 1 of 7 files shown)


moseshll (approved):
I was temporarily confused by the error `"/bin/ls: cannot access '/tmp/stage/toingest/test/*.zip': No such file or directory"` error when the `validate` service runs, but it seems irrelevant to the issue at hand. Since it's not one of the test services. All the other services run without complaint.
My only reservation is the use of MD5 instead of something in the SHA family. Since we're not worrying about a checksum value inside the METS, it looks to me like we could run `sha256sum` at the cost of additional storage. (This is based on the wildly optimistic prediction that MD5 will be fully deprecated in the next ten years.)

aelkiss:
I went with MD5 since this is still really about detecting data corruption rather than malicious attacks. Still, I could potentially see use/value for SHA256 here in that since the backups are being stored in a place we don't really control there's potentially more ability for attackers to change or manipulate the items (encryption notwithstanding). IF we were relying on those checksums for detecting malicious changes we'd also need to back them up somewhere else. My sense is that that level of security isn't really necessary for this material at this time, but worth continuing to think about and potentially revisiting in the future.

PR #22, opened 2020-11-18
HT-2738 virus scan stage for FRAME/EMMA

moseshll:
Not sure about adding `clamav` as a new Docker service. This will probably not run out of the box in a non-Docker environment. The `ClamAV::Client` module that I settled on as sorta supported, does not use the ClamAV API, rather it just talks to the ClamAV daemon and parses the results.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,14 +22,20 @@ services:
       - HTFEED_CONFIG=/usr/local/feed/etc/config_test.yml
       - FEED_HOME=/usr/local/feed
       - TEST=1
+      - CLAMAV_HOST=clamav
+      - CLAMAV_PORT=3310
 
   travis_test:
     build: .
+    volumes:
+      - .:/usr/local/feed
     environment:
       - HTFEED_CONFIG=/usr/local/feed/etc/config_test.yml
       - FEED_HOME=/usr/local/feed
       - TEST=1
-    command: bin/wait-for --timeout=300 mariadb:3306 -- make test TEST_VERBOSE=1
+      - CLAMAV_HOST=clamav
+      - CLAMAV_PORT=3310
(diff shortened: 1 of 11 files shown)


aelkiss (approved):
I think we would need to install `clamav-daemon` on the ingest machines. Shouldn't be a problem.

Tests look good enough for the time being.

aelkiss:
I'll look into what's going on w/ travis here..

docker-compose.yml
```diff
@@ -22,14 +22,20 @@ services:
...
+    volumes:
+      - .:/usr/local/feed
     environment:
       - HTFEED_CONFIG=/usr/local/feed/etc/config_test.yml
       - FEED_HOME=/usr/local/feed
       - TEST=1
-    command: bin/wait-for --timeout=300 mariadb:3306 -- make test TEST_VERBOSE=1
+      - CLAMAV_HOST=clamav
+      - CLAMAV_PORT=3310
+    command: bin/wait-for --timeout=300 mariadb:3306 -- bin/wait-for --timeout=300 clamav:3310 -- make test TEST_VERBOSE=1
```
aelkiss: ah, this is probably not quite working -- I can probably adjust wait-for to take multiple things it can wait for, if it can't already

moseshll: I'll look at the wait-for source. I expected it to need a tweak or two but then it seemed to work.

aelkiss: Already fixed, commit incoming (just need to fix something with the config that's causing the tests to fail)

moseshll: I messed around with using `$self` in the PREMIS config as indicated in the comments (etc/sample_config/00_base_config.yml and etc/config/00_base_config.yml), so I could get the ClamAV version from the daemon directly, but could not get that to work.

PR #24, opened 2020-12-01
HT-2699, HT-2700, HT-2701 - Extract metadata for EMMA objects

aelkiss:
* Adds metadata to EMMA METS (HT-2699)
* Saves metadata in a new database table (HT-2701). This PR also adjusts the database image to use the common HathiTrust database image, which exposed a few bugs around character limits and time zones.
* Fixes some issues with using the virus scan stage in validate_volume.pl etc.


moseshll (approved):
My only issue in running the tests was the one I think has been plaguing me from the beginning: "HTFeed::Namespace::Test is not a Test::Class or an integer at t/classtests.t line 42", if I get rid of it then all tests succeed.

aelkiss:
@respinos  Does the metadata it's extracting & putting in the database work for for the download module for the time being? Any other concerns?

respinos:
@aelkiss yes, that should work.

PR #25, opened 2020-12-09
HT-2523 - Back up material to S3

aelkiss:


aelkiss:
@umjnega You may want to take a look at how this is using awscli to upload & verify material with S3 and make sure it meets your expectations

@moseshll I'm planning to move some of what you had https://github.com/hathitrust/feed/pull/19 in `lib/AWS.pm` with `HTFeed::Storage::S3` from this PR. We could also potentially add an integration test with minio to that.

aelkiss:
It looks like the `ETag` should be the same as what we're storing as Content-MD5, so there's no reason to duplicate it. I'll look into that before merging.

aelkiss:
That said https://aws.amazon.com/premiumsupport/knowledge-center/data-integrity-s3/ does recommend additionally storing it in the metadata. I think it's fine to continue to do that but also to know the `ETag` is available just in case.

aelkiss:
@moseshll  I think this is primarily FYI for future work at this point. I'm going to go ahead and merge and then we can look at some of the follow-up in the new year as time permits.

PR #29, opened 2021-01-22
HT-2748 Remove ingested material from EMMA S3 bucket.

moseshll:
Also work around an apparent issue with S3 `list-objects-v2` on empty bucket returning empty string and confusing JSON parser.

### DIFF
--- a/lib/HTFeed/PackageType/EMMA/Volume.pm
+++ b/lib/HTFeed/PackageType/EMMA/Volume.pm
@@ -31,7 +31,11 @@ sub clean_sip_success {
     bucket => get_config('emma','bucket'),
     awscli => get_config('awscli')
   );
-  $s3->rm('/',"--recursive");
+  my $zip = $self->get_zip_filename;
+  my $xml = $zip;
+  $xml =~ s/\.zip$/.xml/;
+  $s3->rm('/' . $zip);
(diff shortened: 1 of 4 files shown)


lib/HTFeed/PackageType/EMMA/Volume.pm
```diff
@@ -22,6 +23,16 @@ sub move_sip {
...
 
+sub clean_sip_success {
+  my $self = shift;
+
+  $self->SUPER::clean_sip_success();
+  my $s3 = HTFeed::Storage::S3->new(
+    bucket => get_config('emma','bucket'),
+    awscli => get_config('awscli')
+  );
+  $s3->rm('/',"--recursive");
```
aelkiss: I think we only want to remove the actual ingested item on success, rather than everything in the bucket

moseshll: I'd like to use the Volume's `get_mets_filename` and `get_zip_filename` methods to determine the names of the two files to remove from the bucket, but doing so seems to break multiple tests. So, I'm doing something along the lines of EMMA/Volume `move_sip`. Another commit in the works.

moseshll (commented):
I don't know why GitHub is having me review comments on my own PR.

PR #34, opened 2021-02-19
HT-2525 verify cloud storage vs DB

moseshll:


moseshll:
Tests seem to be failing pretty much across the board due to a mariadb connection error.

aelkiss:
I screwed up the database image yesterday; should be fixed now (i'll re-trigger CI). Probably worth thinking about having some kind of CI just for the database image to verify it builds as expected..

lib/HTFeed/StorageAudit.pm
```diff
@@ -0,0 +1,103 @@
...
+                                    awscli => $self->{awscli});
+  my $sql = 'SELECT namespace,id,version FROM feed_backups'.
+            ' WHERE path LIKE "s3://%"'.
+            ' AND deleted IS NULL'.
+            ' ORDER BY lastchecked ASC';
+  $sql .= "LIMIT $limit" if defined $limit;
+  foreach my $row (@{get_dbh()->selectall_arrayref($sql)}) {
+    my ($namespace, $id, $version) = @$row;
+    unless ($s3->s3_has("$namespace.$id.$version.zip")) {
+      $self->set_error('MissingFile', detail => "$namespace.$id.$version.zip not found in AWS");
```
aelkiss: We probably should log this (and other errors) to the `feed_audit_detail` table for reporting (in addition to the regular logging)

aelkiss (approved):
This looks good to me as a start. I would like to try running it on the existing S3 bucket; there are a couple areas I want to improve (like not having to use the 'like' query) that are out of scope of this issue; we should also log to `feed_audit_detail` but I'm OK with saving that until we've verified this works.

lib/HTFeed/StorageAudit.pm
moseshll: Added a second commit with feed_audit_detail logging.

aelkiss (approved):

aelkiss:
Looks good. I'll give it a try and then merge.

aelkiss:
I tried this out; a couple thoughts:

- Checking the presence of each item from the database via awscli is extraordinarily slow. We should probably instead update the `lastchecked` column for each item we find in AWS during the run_not_in_db_check, then report on any items where the `lastchecked` date is before we started running the check.
- While `aws s3api list-objects-v2` is documented to only return 1,000 items at a time, the reality seems to be that if you don't provide `--max-items` it just seems to sit and wait, leading me to believe it may just be trying to retrieve all items in the bucket. Probably best if we explicitly provide a `--max-items` argument; asking for 10,000 items took 4-5 seconds and did indeed return 10,000 items.

aelkiss:
Sorry for the delay on this -- I want to try this again on the existing backup bucket.

aelkiss:
Running it now; I will have a couple small updates.

aelkiss:
Future enhancements to track: 

* save resumption token for awscli for checkpointing / pause / resume
* abstract notion of listing items in storage backend for auditing data den & main repo in the same way instead of having separate audits

aelkiss:
FYI, listing and updating the last checked timestamp for 898,284 HT items (so twice that many objects in S3 for the METS and the zip) took about an hour and 20 minutes using this method, so (in theory) verifying the inventory of the full repository should take about 26 hours, which I think is manageable. Might not want to do it every week, but once a month or so should be fine.

aelkiss:
Ran on the "old" bucket and fixed a few issues; running now on the bucket in the "new" account. I'll merge once that's completed and CI is happy.

aelkiss:
Something's going wrong with respect to the test's abilities to connect to the DB. Could be a mariadb update broke something. I'll review.

aelkiss:
mariadb logs were a red herring; actually this looks like a clamav issue:

```
clamav_1    | + wget -t 5 -T 99999 -O /var/lib/clamav/daily.cvd http://database.clamav.net/daily.cvd
clamav_1    | --2021-03-09 15:17:38--  http://database.clamav.net/daily.cvd
clamav_1    | Resolving database.clamav.net... 104.16.218.84, 104.16.219.84, 2606:4700::6810:db54, ...
clamav_1    | Connecting to database.clamav.net|104.16.218.84|:80... connected.
clamav_1    | HTTP request sent, awaiting response... 403 Forbidden
clamav_1    | 2021-03-09 15:17:38 ERROR 403: Forbidden.
clamav_1    | 
clamav_1    | + wget -t 5 -T 99999 -O /var/lib/clamav/bytecode.cvd http://database.clamav.net/bytecode.cvd
clamav_1    | --2021-03-09 15:17:38--  http://database.clamav.net/bytecode.cvd
clamav_1    | Resolving database.clamav.net... 104.16.219.84, 104.16.218.84, 2606:4700::6810:da54, ...
clamav_1    | Connecting to database.clamav.net|104.16.219.84|:80... connected.
clamav_1    | HTTP request sent, awaiting response... 403 Forbidden
clamav_1    | 2021-03-09 15:17:38 ERROR 403: Forbidden.
```

aelkiss:
Not sure what the issue is other than that freshclam doesn't seem to be able to download the virus definitions. Don't really want to yak-shave this

aelkiss:
There was some external dependency failing for downloading the virus definitions for ClamAV. We could certainly add an integration test for ClamAV at some point, but I don't want the entire suite failing because of that, so for now I've mocked out ClamAV for use in the EMMA tests. I also fixed some issues in the storage audit tests regarding changes I made and needing to run the not-in-AWS check before the not-in-database check runs. If this passes I will go ahead and merge it.

Regarding running in the 'new' account looks like it is working as expected -- it found a number of items that were in AWS but that weren't recorded in feed_backups, which is exactly what we wanted.

PR #36, opened 2021-03-01
HT-2502: Address issues around StorageMigrate and cleaning

aelkiss:
Previously, the behavior of HTFeed::StorageMigrate#clean was untested;
it is called by the job runner but isn't called in the context of
running a particular stage. As things were, this lead to the METS being
removed from the repository after running StorageMigrate.

A second bug with cleaning caused this NOT to happen on success. The
base collate stage attempted to call HTFeed::Volume#clean_sip_success,
which failed for the 'ht' package type. This led cleaning to fail before
it had a chance to remove the METS file.

This addresses the cleaning problems in two ways - one, it overrides the
cleaning behavior in HTFeed::Stage::StorageMigrate to avoid any cleaning
behavior on success or failure, while also testing that the stage does
not leave temporary files behind in staging directories. It also
overrides the cleaning behavior in the
HTFeed::PackageType::HathiTrust::Volume. (This split in cleaning
responsibility doesn't make much sense, but such is the current state of
the current code.) This area is ripe for refactoring when we move to a
job runner such that the cleaning should happen as an integral part of
the stage. Either fix is sufficient to address the problem, but the
belt-and-suspenders approach to help ensure a similar problem doesn't
occur in the future seems worthwhile.

This also addresses the underlying cause of at least one failure, which is
that some old zips in the repository had files such as 'notes.txt' that
were not present in the METS files. Storage migration should not fail
because of this (although we may want to audit for volumes like this at
some point in the future)


aelkiss:
FYI @moseshll you may find the tests for storage migration here interesting -- we needed to create a stage to copy existing things from the repository to Glacier Deep Archive and Data Den; running it in production exposed some problems outside the scope of the stage. Worth thinking about from a refactoring perspective later on - I think long term the path looks like getting ingest working in Docker, getting the individual stages running under some other batch job scheduling thing rather than our own home-grown buggy thing, and (ideally) doing some substantial refactoring with Volume. That all should end up with a more maintainable code base without needing to totally rewrite it in Ruby etc up-front.

PR #37, opened 2021-03-12
Audit Data Den zipfiles against DB and METS.

moseshll:
Leaving this as draft, because I am not so sure about the `DataDenZipAudit` initializer grabbing the right kind of `HTFeed::Storage` object in its initializer. (See the "EVIL HACK" line referencing `HTFeed::Storage::for_volume()`)


lib/HTFeed/DataDenZipAudit.pm
```diff
@@ -0,0 +1,169 @@
...
+  my $path_prefix = shift || '/htdataden';
+
+  my $sql = 'SELECT COUNT(*) FROM feed_backups' .
+            " WHERE path LIKE '$path_prefix%'" .
+            ' AND deleted IS NULL AND saved_md5sum IS NOT NULL';
+  my $dbh = HTFeed::DBTools->get_dbh();
+  my ($rows) = $dbh->selectrow_array($sql);
+  my $random_row = int(rand($rows));
+  my ($namespace, $objid, $path, $version);
+  $sql = 'SELECT namespace,id,path,version FROM feed_backups' .
```
aelkiss: I believe we can do `order by rand() limit 1` -- the approach here might be faster, but I think probably not worth the increase in complexity in this case

lib/HTFeed/DataDenZipAudit.pm
```diff
@@ -0,0 +1,169 @@
...
+  $object->{zip_path} = $object->{path} . '/' . $object->{objid} . '.zip.gpg';
+  
+  my $volume = HTFeed::Volume->new(
+    packagetype => 'pkgtype',
+    namespace   => $object->{namespace},
+    objid       => $object->{objid}
+  );
+  $object->{volume} = $volume;
+  # EVIL HACK that may break in production.
+  $object->{storage} ||= (HTFeed::Storage::for_volume($volume))[0];
```
aelkiss: Yeah, I think what we need here is a name for the particular instance of the storage back end, both in the config and in the database, so we can instantiate the correct one. There is a ticket (https://tools.lib.umich.edu/jira/browse/HT-2812) for that -- perhaps we should queue that up for next sprint?

lib/HTFeed/DataDenZipAudit.pm
```diff
@@ -0,0 +1,169 @@
...
+    die "Database query failed: $@";
+  }
+  return { namespace => $namespace, objid => $objid,
+           path => $path, version => $version };
+}
+
+sub new {
+  my $class = shift;
+
+  my $object = {
```
aelkiss: Generally in this code base I've used `$self` for the object we build in the constructor.

lib/HTFeed/DataDenZipAudit.pm
moseshll: Works for me. I'm gonna leave this part of the code alone for now since it does work with its rickety test scaffolding.

aelkiss (changes requested):
Looks like this one still needs the `storage_name` update / "evil hack" excision

moseshll:
What a mess, thanks to my rebase. You want a clean branch/PR? It's just 3 files.

aelkiss:
Superceded by #49

PR #44, opened 2021-04-23
Change layout for versioned backups

aelkiss:
- inode usage for pairtree in backup storage was too high (on average,
created several directories per actual file)

- New scheme: paths only one deep under namespace, named using a prefix of
the object ID of length max(3, length(objid-4)). This was empirically
determined to result in an mean directory size of around 80 objects,
median of 13, and maximum of around 30,000. We have been told that
performance should not suffer dramatically until we have hundreds of
thousands or millions of files in a single directory. This should result
in an overhead of only about 200,000 directories for ~34 million actual
files.

- This also eliminates the version directory and the individual object
directory

- Refactors Storage to add methods for determining the zip & METS
filename within the context of the storage rather than assuming they are
the same as the filenames of the original object

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,6 +2,18 @@ version: '3'
 
 services:
 
+  test:
+    build: .
+    volumes:
+      - .:/usr/local/feed
+    environment:
+      - HTFEED_CONFIG=/usr/local/feed/etc/config_test.yml
+      - FEED_HOME=/usr/local/feed
+      - TEST=1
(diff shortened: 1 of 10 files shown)


aelkiss:
@moseshll  FYI, if you have time to take a look. The motivation here was that with the pairtree structure we have the inode usage was very high compared to the actual number of files we had... we were requested for the data den backup to try to reduce inode usage so much as possible. I experimented with a number of options and found this layout to work the best of the ones I tried.

PR #46, opened 2021-05-12
HT-2812 Add name for storage back-end to config & record in feed_backups

moseshll:


t/storage_s3.t
```diff
@@ -1,5 +1,6 @@
 use Test::Spec;
 use Test::Exception;
+use HTFeed::Log {root_logger => 'INFO, screen'};
```
aelkiss: `use HTFeed::Test::SpecSupport` instead, which will set up a logger

aelkiss (commented):
:+1:  Generally looks great; one small change to logger setup in `storage_s3.t` but otherwise looks exactly like what I was looking for.

aelkiss:
Looks great. The config needs to be updated when this is deployed, so I'll do all that next week. Once that's merged we should be able to address the concerns on the other two PRs and get those merged as well.

aelkiss (approved):

PR #47, opened 2021-05-24
HT-2516 Data Den expiration

moseshll:
This is just a draft. I'd be interested if I'm barking up the wrong tree. Production hasn't been updated with `feed_backups.storage_name` yet, so unsure whether to use that or the `path LIKE` clause that's there now.


aelkiss:
Let's use `feed_backups.storage_name` -- I'm waiting for a process that's moving some stuff around in data den to complete before I deploy that. Hopefully before Friday... once that is sorted we should be able to move forward quickly with some of this other stuff. I'll take a look at the draft PR today.

bin/expire_dataden.pl
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+  unless (defined $volume_storage) {
+    die "Unable to get storage '$row->{storage_name}' for $volume->{namespace}.$volume->{objid}";
+  }
+  my $update_sth = get_dbh()->prepare(<<'SQL');
+      UPDATE feed_backups SET deleted=1
+      WHERE namespace=? AND id=? AND version=? AND storage_name=?
+SQL
+  unless (unlink $volume_storage->mets_filename, $volume_storage->zip_filename) {
```
aelkiss: We should probably move this logic into the individual storage providers, since it will differ for Data Den vs. S3. I think that is the only thing that will differ, though.

aelkiss (commented):
The overall direction looks good.

I think it is worth testing, probably with a mocked storage provider, since the SQL query is complicated and the consequences if it misbehaves are potentially dire - basically load some data to the feed_backups table and test a few scenarios:

* single version is retained
* multiple versions where all versions are < 6 months old are retained
* multiple versions where at least one version is < 6 months is removed

We should also log when we remove versions.

bin/expire_dataden.pl
aelkiss: Yes, I think that makes sense -- not so much because the base `Storage` object *couldn't* provide a reasonable default implementation (i.e. what you have here), but because I think deletion *shouldn't* be supported by default for all storage backends.

moseshll:
This should get us a ways toward having HT-2526 supported as well, just a matter of adding the script and the tests. Are you okay with having the storage name hardcoded in the `bin` script, e.g., `HTFeed::BackupExpiration->new(storage_name => 'dataden');`

aelkiss:
Let's take the storage name as a command-line argument.

moseshll:
Makes sense. I'll add the command-line arg tomorrow.

aelkiss:
Superceded by #48

PR #48, opened 2021-06-10
HT-2516, HT-2526 - expiration for backup storage

moseshll:


moseshll:
This is branched off of HT-2516 so it's a little weird. I don't normally do that.

aelkiss:
No worries. I'll just review them both together and mark #47  as superceded.

t/backup_expiration.t
```diff
@@ -0,0 +1,250 @@
...
+          ok(!-e $mets, "old ($old_storage->{timestamp}) mets deleted");
+          ok(!-e $zip, "old ($old_storage->{timestamp}) zip deleted");
+        }
+        my $mets = $storage->mets_obj_path();
+        my $zip = $storage->zip_obj_path();
+        ok(-e $mets, "new ($new_version) mets left intact");
+        ok(-e $zip, "new ($new_version) zip left intact");
+      };
+
+      it "does not do anything with old versions if there is no new one" => sub {
```
aelkiss: I'm not sure this (and the one at line 223) are quite right - in this scenario, I would expect this could happen if we hadn't run the expiration/cleanup script recently, and the expected behavior would be that the oldest version was pruned.

I think it's also worth testing what happens if we have two versions, *neither* of which is older than 6 months -- in this case it should leave both new versions intact.

t/backup_expiration.t
```diff
@@ -0,0 +1,250 @@
...
+      it "succeeds" => sub {
+        my $exp = HTFeed::BackupExpiration->new(storage_name => 'prefixedversions-test');
+        ok($exp, 'new returns a value');
+      };
+    };
+
+    describe "#run" => sub {
+      it "does not do anything with a single version" => sub {
+        my $storage = staged_volume_storage('prefixedversions-test', 'test','test');
+        my $version = old_random_timestamp();
```
aelkiss: This sequence shows up several times in these tests (as well as for objectstore where `stage`/`make_object_path` is a no-op); I would consider extracting these to a helper method and perhaps seeing if there's any other repeated setup in these tests that might be clearer or easier to maintain if we extracted it

lib/HTFeed/BackupExpiration.pm
```diff
@@ -0,0 +1,93 @@
...
+  # volumes, nothing more granular.
+  my $sth = get_dbh()->prepare(<<'SQL');
+    SELECT namespace,id,MIN(version) AS min_version,MAX(version) AS max_version
+    FROM feed_backups
+    WHERE deleted IS NULL
+      AND storage_name=?
+    GROUP BY namespace,id
+    HAVING COUNT(*) > 1
+      AND min_version < DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
+      AND max_version > DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
```
aelkiss: See comment on tests below, but I'm not sure this will work as expected if we (say) ingest the item twice on the same day, or if for whatever reason we go for a while without running this expiration script. The desired behavior is that if it has two versions that are both older than 6 months, only one is retained.

aelkiss: So I think we want to check for:

* at least two versions
* at least one of them is older than 6 months

i.e. I think we can drop the `min_version` clause

lib/HTFeed/BackupExpiration.pm
```diff
@@ -0,0 +1,93 @@
...
+    # Get the versions in this group that can be deleted.
+    # The previous query has verified that there is >= 1 unexpired version.
+    my $versions_sth = get_dbh()->prepare(<<'SQL');
+      SELECT version
+      FROM feed_backups
+      WHERE deleted IS NULL
+        AND storage_name=?
+        AND namespace=?
+        AND id=?
+        AND version < DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
```
aelkiss, line 57: This will need adjustment given the previous comment -- I'm not sure what the easiest way to express that is, but I think we want to:

- in no case should we delete the most recent version (even if it is more than 6 months old)
- in no case should we delete anything newer than 6 months old

aelkiss (changes requested):
See comments in-line. In thinking it through I think the logic isn't quite as desired. Seeing the tests helped me think through that. I think it would also be nice if possible to factor out some of the repetitive setup in the tests.

aelkiss:
Saw the request for re-review; I was on vacation last week and may not get to this until later this week.

aelkiss (approved):
Logic looks good now, and the tests are well-specified in terms of covering the cases we're interested in.

It looks like the shape of the tests between `ObjectStore` and `PrefixedVersions` are the same now. That makes me think that we should probably consider making this into a true unit test for `HTFeed::BackupExpiration` by using mocked storage like here: https://github.com/hathitrust/feed/blob/master/t/collate.t#L13 and make sure that we test deleting things in the storage backend tests. Alternatively we could use the shared example group functionality from `Test::Spec` to test both -- https://metacpan.org/pod/Test::Spec#Shared-example-groups

I'll go ahead and merge this now as I think this completes HT-2516 and HT-2526. We can continue some of the refactoring on HT-2866.

PR #51, opened 2021-07-07
HT-3034 Ensure storage classes get loaded by zip audit

moseshll:
A more dynamic approach to zip audit storage classes.

Also adds a potentially useful amount of logging (HT-3035), not sure if it is sufficient at this point, though.
May be first step in umbrella refactoring ticket HT-2866.


moseshll:
I should add that StorageZipAudit.pm is meant to merge into StorageAudit.pm at some point, when the disjoint APIs can be reconciled with each other.

aelkiss:
Unfortunately I won't have a chance to review this until next Monday the 12th, but I will make it a priority to do so then.

lib/HTFeed/Storage/ObjectStore.pm
```diff
@@ -201,5 +212,10 @@ sub record_backup {
   return $rval;
 }
 
+sub audit_path {
+  my $self = shift;
+
+  return "s3://$self->{s3}{bucket}/" . $self->object_path;
```
aelkiss, line 232: Based on the need in HT-3057 we may end up just putting this in `object_path`, but I think this is correct for right now.

lib/HTFeed/StorageZipAudit.pm
```diff
@@ -0,0 +1,199 @@
...
+  }
+  my $obj = {namespace => $namespace, objid => $objid, version => $version,
+             path => $path, storage_name => $self->{storage_name}};
+  $obj->{volume} = HTFeed::Volume->new(packagetype => 'pkgtype', namespace => $namespace, objid => $objid);
+  my $config = get_config('storage_classes');
+  $obj->{storage_config} = $config->{$self->{storage_name}};
+  $obj->{storage} = $obj->{storage_config}->{class}->new(volume => $obj->{volume},
+                                                         config => $obj->{storage_config},
+                                                         name   => $self->{storage_name});
+  $obj->{storage}->{zip_suffix} = '.gpg';
```
aelkiss: The handling around encryption is a little bit messy to begin with and might be worth rethinking -- definitely would be preferable for the storage backend to know whether or not objects were supposed to be encrypted when restoring and use the right suffix. Low priority.

lib/HTFeed/StorageZipAudit.pm
```diff
@@ -0,0 +1,199 @@
...
+  $obj->{storage}->{timestamp} = $obj->{version};
```
aelkiss: Perhaps the storage backend constructor should take the timestamp as a parameter? Medium priority.

lib/HTFeed/StorageZipAudit.pm
```diff
@@ -0,0 +1,199 @@
...
+  $obj->{zip_file} = $obj->{storage}->zip_filename();
+  $obj->{mets_file} = $obj->{storage}->mets_filename();
+  $obj->{zip_path} = $self->storage_object_path($obj) . '/' . $obj->{zip_file};
```
aelkiss: Hmmm.. Perhaps we should consider using `$obj->{storage}->zip_obj_path` / `$obj->{storage}->mets_obj_path` and asserting that those are the same as what's in the database table? Maybe kind of medium priority?

lib/HTFeed/StorageZipAudit.pm
```diff
@@ -0,0 +1,199 @@
...
+  my $error_count = 0;
+  my $objects = (defined $obj) ? [$obj] : $self->all_objects();
+  unless (0 < scalar @$objects) {
+    get_logger->trace("no auditable files: returning");
+    return $error_count;
+  }
+  foreach my $obj (@$objects) {
+    eval {
+      die "encrypted zip $obj->{zip_path} does not exist" unless -f $obj->{zip_path};
+      my $db_ok = $self->check_encrypted_zip_against_database($obj);
```
aelkiss: I think it's fine for now to assume the zip is encrypted, but we wouldn't be able to make that assumption if we use this for auditing the primary repository as well. Low priority to change right now.

lib/HTFeed/StorageZipAudit/ObjectStore.pm
```diff
@@ -0,0 +1,115 @@
...
+  if ($@) {
+    die "Database query failed: $@";
+  }
+  return $self->{pending_objects};
+}
+
+# Request a single object from Glacier. To keep storage costs down,
+# use the restore_request_issued flag to keep the requests under control
+# in case something else goes haywire.
+sub _request_object {
```
aelkiss, line 52: I wonder whether this, `_restore_object`, and `_check_object` ought to be in `HTFeed::Storage::S3` (or perhaps `HTFeed::Storage::S3::Glacier`)? Medium priority.

lib/HTFeed/StorageZipAudit.pm
```diff
@@ -0,0 +1,199 @@
...
+  my $class        = shift;
+  my $storage_name = shift;
+
+  if (defined $storage_name) {
+    # Instantiate audit subclass based on storage class name.
+    my $config = get_config('storage_classes');
+    my $storage_config = $config->{$storage_name};
+    die "unable to get config for $storage_name" unless $storage_config;
+
+    my @components = split '::', $storage_config->{class};
```
aelkiss: My first thought here was that this is a little bit hairy, and it might be preferable to explicitly specify in the storage config what audit class to use. However, it also seems unlikely that we would want to vary the audit class independently of the storage class. Perhaps we could instead add a method on the storage instance that returns a zip auditor for that storage? Then we would start the audit as something like `HTFeed::Storage::storage_for('instance')->zip_auditor->run()`

aelkiss: That would also allow us to get rid of the empty `HTFeed::StorageZipAudit::PrefixedVersions` subclass

lib/HTFeed/StorageZipAudit/ObjectStore.pm
```diff
@@ -0,0 +1,115 @@
...
+  my $self = shift;
+  my $obj  = shift;
+
+  unless ($obj->{tmpdir}) {
+    $obj->{tmpdir} = File::Temp->newdir();
+  }
+  return $obj->{tmpdir};
+}
+
+sub all_objects {
```
aelkiss, line 26: I like the approach here of restoring restorable objects and then essentially making this act like any other storage.

lib/HTFeed/StorageZipAudit.pm
```diff
@@ -0,0 +1,199 @@
...
+
+  my $error_count = 0;
+  my $objects = (defined $obj) ? [$obj] : $self->all_objects();
+  unless (0 < scalar @$objects) {
+    get_logger->trace("no auditable files: returning");
+    return $error_count;
+  }
+  foreach my $obj (@$objects) {
+    eval {
+      die "encrypted zip $obj->{zip_path} does not exist" unless -f $obj->{zip_path};
```
aelkiss, line 123: I'd like to see an info-level logging call here to indicate that we have started a check for a particular item. High priority.

t/storage_zip_audit.t
```diff
@@ -0,0 +1,232 @@
...
+        $sql = 'SELECT COUNT(*) FROM feed_audit_detail' . 
+               ' WHERE namespace = ? AND id = ? AND status = "BadChecksum"';
+        my @res = get_dbh->selectrow_array($sql, undef, 'test', 'test');
+        is($res[0], 1, 'BadChecksum error recorded');
+      };
+    };
+  };
+
+  describe "HTFeed::StorageZipAudit::ObjectStore" => sub {
+    describe "#new" => sub {
```
aelkiss: The `Test::Spec` [shared_examples_for](https://metacpan.org/pod/Test::Spec#shared_examples_for-DESCRIPTION-=%3E-CODE) feature combined with the `around` hook for setting the `S3_HEAD_OBJECT_RESTORE_DONE` environment variables might be worth looking at for the things shared between groups. For the things that vary between examples (primarily the name of the storage and the storage class) you could look at the [share %HASH](https://metacpan.org/pod/Test::Spec#share-%HASH) feature of `Test::Spec`. Medium priority.

aelkiss (commented):
Looking good so far. I made a number of comments on possible future directions inline. I'll call out which ones I think we should prioritize, though.

lib/HTFeed/StorageZipAudit.pm
aelkiss: I think this change should be high priority

lib/HTFeed/StorageZipAudit.pm
moseshll: Well, Storage uses the presence of $self->{config}{encryption_key}, it's not even really up to the subclass at collation time. But relying on that seems brittle too. I don't like the plethora of `xxx_path()` methods but maybe it's time for something like `storage_zip_path` that's explicitly about working from the "destination" end of the process. I'm going to let it lie for now.

aelkiss: :+1:

lib/HTFeed/StorageZipAudit.pm
moseshll: I may be misunderstanding your comment, but this path juggling is specifically for the Glacier case where we make a temp location for the downloaded files. I mean, we could make that assertion for any class other than ObjectStore. Maybe I should call it ephemeral_object_path.

lib/HTFeed/StorageZipAudit/ObjectStore.pm
moseshll: I'd separate out the SQL `feed_backups` updates, however. So restore request made by some other process doesn't end up as unwilling audit participant.

lib/HTFeed/StorageZipAudit.pm
aelkiss: Right -- I think I didn't quite understand that at first glance here,  and was thinking primarily about the Data Den case. It's probably worth thinking about how to isolate the special behavior to ObjectStore, but I think that's pretty low priority for right now.

lib/HTFeed/StorageZipAudit/ObjectStore.pm
aelkiss: :+1:

lib/HTFeed/StorageZipAudit.pm
moseshll: I can get rid of the redundant `zip_file` and `mets_file` attrs regardless.

moseshll: The way I rationalize it, the storage class need not know where to "download" things. ObjectStore is just weird.

lib/HTFeed/StorageZipAudit.pm
moseshll: I'm introducing `encrypted_by_default` and `set_encrypted` methods on `Storage`. Flawed they may be, but better than poking around inside the storage attrs.

lib/HTFeed/StorageZipAudit.pm
moseshll: I'm replacing the call to storage's `validate_zip_checksum` with `validate_zip` so it will be encryption agnostic going forward.

bin/storage_zip_audit.pl
```diff
@@ -0,0 +1,16 @@
...
+use warnings;
+use FindBin;
+use lib "$FindBin::Bin/../lib";
+use HTFeed::StorageZipAudit;
+use HTFeed::Log {root_logger => 'INFO, screen'};
+
+die "Specify a feed_backups.storage_name value" unless 1 == scalar @ARGV;
+my $storage_name = $ARGV[0];
+
+my $audit = HTFeed::StorageZipAudit->new($storage_name);
```
aelkiss: I think we still need to adjust this to get the storage and ask for the zip auditor rather than constructing StorageZipAudit directly.

lib/HTFeed/StorageZipAudit.pm
```diff
@@ -0,0 +1,196 @@
...
+  unless ($namespace && $objid && $path && $version && $md5) {
+    croak "invalid storage_object() args: namespace, objid, path, version, md5 required";
+  }
+  my $obj = {namespace => $namespace, objid => $objid, version => $version,
+             path => $path, storage_name => $self->{storage_name},
+             md5 => $md5};
+  $obj->{volume} = HTFeed::Volume->new(packagetype => 'pkgtype', namespace => $namespace, objid => $objid);
+  my $config = get_config('storage_classes');
+  $obj->{storage_config} = $config->{$self->{storage_name}};
+  $obj->{storage} = $obj->{storage_config}->{class}->new(volume    => $obj->{volume},
```
aelkiss, line 91: Blargh. This ends up more complicated than I'd like, since storage is specific to a particular volume, but the zip audit isn't. I think there is a missing concept here of the overall storage instance as opposed to the storage for a particular volume (or rather, that concept only exists via name rather than as a particular class instance.) I don't think that's worth addressing now, but I made a note on HT-2521.

bin/storage_zip_audit.pl
aelkiss: Except we can't, because the storage needs to be bound to a particular volume, which we don't have right now... I'm not sure I have an elegant solution for getting the right zip audit class right now. Maybe we just specify it in the config per storage back end for the time being.

aelkiss (changes requested):
See comments inline; most of the changes look good. In particular I'm happy the shared examples in the tests worked out.

However, I think I screwed up in suggesting that the storage class know the zip auditor, since I had forgotten the instance of the storage class was bound to a particular volume. We might need to back away from that and see about adding the concept of a storage class independent of a volume later on. I added some comments on that in-line.

lib/HTFeed/StorageZipAudit.pm
aelkiss: Another option could be splitting up `ZipAuditor` into two pieces. One piece would take the storage name, pick a volume using that storage, and fetch info from the database about the volume. That's independent of the particular storage backend. The second piece, which could be bound to a volume and returned by the storage backend, would do the actual verification work. That might be easier to do now without big changes to `Storage`.

moseshll: I've gone ahead and worked a bit on merging the zip audits with the filesystem crawls. I'm implementing iterators, for example, as Storage class methods rather than going with the route of needing a storage instance bound to a volume. Until we're actually talking about volumes, that is. It's maybe too early to tell if that's going to work well, but I may be able to scrape away some of the complexity.

In the case you cite, I've moved storage_config wrangling rom the storage_object to the StorageZipAudit initializer.

Some of the path parsing functionality from audit/DataDenPath.pm looks like a good fit for PrefixedVersions.pm as one or more class methods.

bin/storage_zip_audit.pl
moseshll: (If only I had remembered to update that file, I might have seen the potential complication! Alas, egg is on my face.)
I actually think the simplest thing, from where I'm sitting, would be to have zip_auditor as a `Storage` class method. Right now in the tests we're shoveling storages all over the place so it's natural to query the instance about an auditor. In terms of starting from a blank slate, I see some fun approaches:

1. Make zip_auditor a Storage class method. Storage subclass need know only the existence of a certain other class.
2. Shove auditor class into config. Replacing spaghetti code with spaghetti config (I am increasingly not a fan of this  seemingly industry-wide push toward increasingly minilanguage-like "configuration goo" [a phrase I have seemingly stolen from Doug McIlroy]).
3. Resurrect my ham-handed "class cluster" initializer from PR part *un*. (Some kind of cluster anyway.)
4. Determine the auditor class dynamically based on the storage class. This has a certain self-contained meta charm but sounds like it could require the StorageZipAudit class to know too much "stuff".

I think I'm going to take a crack at option 1 as I sense Perl fatigue creeping up on me.

aelkiss (approved):
This looks great -- thanks.

aelkiss:
@moseshll  Is this ready to merge at this point?

moseshll:
@aelkiss Yes, probably wise to merge now.

I will probably have the next commits under the umbrella of HT-2866.

PR #52, opened 2021-07-14
HT-3057: use s3 cp instead of s3api put-object

aelkiss:
s3 cp automatically uses multi-part uploads and should generally have
better performance:
https://aws.amazon.com/premiumsupport/knowledge-center/s3-multipart-upload-cli/

One drawback is that we cannot specify the md5 checksum to validate
against with s3 cp:
https://aws.amazon.com/premiumsupport/knowledge-center/data-integrity-s3/

However, s3 cp will still calculate the checksum internally and compare
it before accepting the upload:
https://docs.aws.amazon.com/cli/latest/topic/s3-faq.html

There is potentially a race condition if the object is modified after we
compute the checksum but before s3 cp computes the checksum. We also
can't easily test/verify the documented behavior from our side the way
we can with the --checksum-md5 option to put-object. Ultimately, this
means we need a little bit more trust that aws cp does what the
documentation says. The other option (manually doing our own multi-part
uploads a la the multipart documentation above) seems even less
preferable.

### DIFF
--- a/lib/HTFeed/Storage/ObjectStore.pm
+++ b/lib/HTFeed/Storage/ObjectStore.pm
@@ -136,23 +136,21 @@ sub postvalidate {
 
 sub move {
   my $self = shift;
-  $self->put_object($self->mets_key,$self->{mets_source});
-  $self->put_object($self->zip_key,$self->{zip_source});
+  $self->cp_to($self->{mets_source},$self->mets_key);
+  $self->cp_to($self->{zip_source},$self->zip_key);
 }
 
-sub put_object {
+sub cp_to {
   my $self = shift;
-  my $key = shift;
   my $source = shift;
+  my $key = shift;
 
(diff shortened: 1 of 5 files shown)


aelkiss:
@moseshll This was motivated by repeated failures to upload moderately large (8+ GB) objects to S3 using `s3api put-object`; we'd get messages like:

`awscli failed with status 65280, error \n('Connection aborted.', BrokenPipeError(32, 'Broken pipe'))\n`

Mainly I'm hoping you can review the AWS documentation and see if my conclusions make sense, and make sure that the changes to the code & tests make sense given all that. No need to run the tests locally (assuming that CI passes) unless you want to.

moseshll (approved):
I agree with your judgment on the Amazon documentation. The low-level multipart upload procedure sounds like a nightmare, so I'd say good to go.

PR #54, opened 2021-07-20
Ht 2866 backup management

moseshll:


aelkiss:
Looking at the failure -- I'm not offended by the completeness check returning 2 errors if there is both a zip and METS in storage that don't match something in the database, but a couple other thoughts:

- I didn't see where `is_object_zip_in_storage` / `is_object_mets_in_storage` are being called, but it would probably be preferable to avoid that extra call to AWS if we're listing all the objects anyway. I *think* with AWS we even get things back from the list call in a sorted order by default, so it should be possible to ensure that for each thing we get back from the list there's both a METS and a ZIP without needing to generate the whole list first.

- For Data Den, we probably want to get all the entries in a given directory and sort them for that same reason (although determining whether a particular path is on disk is not as expensive as with AWS). `find` might not be as useful as traversing the path hierarchy ourselves. `find | sort` would probably work but may be too slow and brittle. `find > temp_file; sort temp_file` might be preferable, although we'd still need to wait for the entire find & sort to complete before doing any other work, whereas if we traversed the directory structure inside the program and sorted the contents of each directory we could at least start updating the database sooner. That said, there isn't any particular rush to complete an audit, and I don't know if the extra development work would be worth it right now.

moseshll:
Okay. The current tests should pass if I back off on requiring deduplication succeed. Let me address your added points:

- The two "is in storage" routines are called from the storage completeness check so the thing we're iterating is the DB table. I have a note that of the two completeness routines, `run_storage_completeness_check` should be called second since it allows `lastchecked` to be set on the auditor. In theory, if the storage completeness check is run second, it should be picking up the leftovers. It's an empirical question just how massive that batch of leftovers could be. A few thousand? I guess that all depends on the circumstances that could leave orphans in `feed_backups`. (However, I did just realize that by only running the storage completeness task on a subset of storage, I have jettisoned ability to check that both METS and zip are present. As you say, AWS is going to give us a sorted list, so I need to put a check on iterator returns for both file types in `run_database_completeness_check`.)
- I would much rather write a directory traversal than mess with `find | sort > file`. If only we had the BSD/OS X `find -s` at our disposal! I assume this is the approach I should take when I address crawling the full repo. I note that `audit/fsCrawl.pl` seems to be prone to the same `$prevpath` issue with unsorted `find` output.

So my to-do list is:
1. Back out debugging code in my second commit and relax the iterator tests.
2. Save some state in `run_database_completeness_check` that allows it to verify zip and METS both present.
3. `find` replacement for Data Den iterator.

aelkiss:
Regarding "is in storage" -- yes, this seems fine if we're only calling it for things in the database table that seem to be leftovers. We would never expect there to be any of those; there are no normal circumstances that would lead to this apart from accidental or intentional changes made to the storage backend outside of the ingest process. It would be far more likely to find something in storage that is not in the database table -- e.g. it should be possible for the storage upload process to complete, but some further error prevents that being recorded; it should not be possible for ingest to record that upload completed without it actually having done so.

Regarding unsorted find output - that shouldn't be an issue for the 'regular' repository, since find will always print everything in the same terminal directory together, and there is only one HT item in each terminal directory.

Regardless the to-do list sounds good to me.

aelkiss (approved):
Looks good as far as I can tell. Do you think we're ready to try running this on the real repository?

moseshll:
I think it should be worth a try. We know there will be some fine tuning.

PR #55, opened 2021-07-23
Build packages to GHCR including private submodules

aelkiss:


aelkiss:
FYI @daaang , @moseshll -- this helps unblock a messy situation where we have been running ingest and backup migration partially in pods Kubernetes, but using code deployed using the old (Phil) deploy process rather than from a container. The reason for that was because we need to have these private submodules because some of the code there may be covered by NDA. With GitHub Container Registry, we can build both a public version (for use by partners, etc) that does not include the NDA code as well as a full version that does, while scoping the internal version to just HathiTrust.

@daaang  I may need help making it so that Kubernetes is able to access internal GHCR packages, but we can cross that bridge when we get there.

PR #56, opened 2021-08-02
Refactor tests from t/backup_expiration.t

moseshll:
Using the shared example feature (option two in the ticket description). Do we still need a mocked storage approach a la collate.t?

### DIFF
--- a/t/backup_expiration.t
+++ b/t/backup_expiration.t
@@ -53,6 +53,7 @@ describe "HTFeed::BackupExpiration" => sub {
     $storage->make_object_path;
     $storage->move;
     $storage->record_backup;
+    $storage->cleanup;
     return $storage;
   }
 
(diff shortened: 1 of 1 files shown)


aelkiss:
Sorry I haven't gotten to taking a look at this one -- I will tomorrow or Friday morning.

aelkiss (approved):
The shared examples look great to me.

I do wonder if perhaps for `zip_deleted` and `mets_deleted` we could use `HTFeed::StorageAudit::is_object_{zip,mets}_in_storage` rather than the conditional here. On the one hand, the test does check concrete state rather than relying on implementation, and it limits dependencies to just `HTFeed::Storage`. On the other hand, it is a little bit messy.

I suppose another option would be to move the specific implementation for `zip_deleted` and `mets_deleted` to a callback in the `describe` blocks for the specific storage; that would still have the duplicate implementation but would at least put the storage-class-specific part of the test in a place that makes more sense.

That all said, (as usual) I think it's fine to leave this as-is for the time being, but keep these options in mind and revisit if 1) we need to make changes to these tests or 2) we do some of the refactoring at HT-2519 or HT-2521 that could move some of these functions for actually getting an object from storage into the class under test.

PR #59, opened 2021-08-16
Prepare for running ingest fully from container

aelkiss:
- update dependencies to debian 11 (bullseye)
- don't run container as root
- clean up configuration & prepare for config & secrets injection from kubernetes
- get version from environment variable (by default github commit used to build the docker image). Could inject the tag when we do a proper release.


.github/workflows/build-ht-3019.yml
```diff
@@ -0,0 +1,60 @@
+name: Docker Build
```
aelkiss: To remove before merging

aelkiss:
See also https://github.com/hathitrust/ht_tanka/pull/3

aelkiss:
A lot of this is cleaning up historical stuff that should have been cleaned up a while ago. The main questions for review are whether these look good:

- Dockerfile(s)
- github actions
- does config strategy look reasonable (what's in this repo vs. in tanka vs. injected via secrets)

aelkiss:
CI is all failing because of issues with `test-tmp` (which we should probably not be creating under `t` anyway). Will look into that tomorrow.

aelkiss:
n.b. @botimer looked at this with me and thought the Dockerfile and github actions stuff looked fine.

PR #65, opened 2021-09-15
HT-3018: End-to-end testing for ingest & audit in dev

aelkiss:
- Fix warnings from fsCrawl; rename to main_repo_audit to better reflect
purpose

- remove start_audit.sh (since we will manage this with kubernetes)

- Add handle table for development (no handle query support, but the
handle stage can work this way)

- ingest_ia_dev.pl; ingests a single IA item end-to-end for
testing, and backdates it so we can ensure it gets checked via
auditing

There are still some sharp edges:

- The permissions on the repository directories will be root by default;
those have to be manually changed to the ingest user (see note in
docker-compose.yml)


aelkiss:
FYI @moseshll  -- I'm trying to get to the point where the audit stuff is merged back into the main repository and we can run it on a routine basis.

aelkiss:
It would probably be preferable to use a real handle server in a docker image -- e.g. https://github.com/datacite/docker-handle, https://hub.docker.com/r/osul/handle -- and to update it via the REST API -- but we aren't quite there yet (in particular I don't know if we ever got updates via the REST API working in production)

aelkiss:
md5 audit is running & working as expected.

PR #67, opened 2021-09-20
HT-3181: Fixes for backup auditing

aelkiss:
- test that lastmd5check is updated on audit
- fix errors in error formatting
- use encryption config from storage (don't assume specific encryption
types are always encrypted or not -- breaks testing in dev w/o
encryption)
- change all_objects -> objects_to_audit
- add flags for storage_audit.pl to run checksum or inventory audit
- move backup audit script into audit dir & rename to backup audit
(eventually, should be able to audit the full repo as well, but for now
that's separate)
- autoload storage classes in for_storage


moseshll (approved):
This looks good to me. I wasn't able to get all tests to pass locally, for whatever reason, but the testing issues look to be unrelated to the changes made here.

PR #70, opened 2021-09-29
HT-3142 Automatically download material from Dropbox at ingest time

moseshll:
WIP for dropbox download.

### DIFF
--- a/lib/HTFeed/PackageType/Simple/Download.pm
+++ b/lib/HTFeed/PackageType/Simple/Download.pm
@@ -1,0 +1,73 @@
+package HTFeed::PackageType::Simple::Download;
+
+use warnings;
+use strict;
+use base qw(HTFeed::Stage::Download);
+use HTFeed::Config qw(get_config);
+use Log::Log4perl qw(get_logger);
+
+my $download_base = "dropbox:";
+
+sub run {
+  my $self = shift;
+
+  my $volume = $self->{volume};
+  # check if the item already exists at $volume->get_sip_location()
+  my $sip_loc = $volume->get_sip_location();
+  if (-f $sip_loc) {
+    get_logger->trace("$sip_loc already exists");
+  } else {
+    my $rclone_config_path = get_config('rclone_config_path');
+    unless (defined $rclone_config_path) {
+      $self->set_error('StageFailed', detail => "rclone_config_path not configured");
+      return 0;
+    }
+    my $nspkg = $volume->get_nspkg();
+    my $dropbox_folder;
+    eval { 
+      $dropbox_folder = $nspkg->get('dropbox_folder');
+    };
+    if ($@) {
+      $self->set_error('StageFailed', detail => $@);
+      return $self->succeeded();
+    }
+    my $filename = $volume->get_SIP_filename();
+    my $url = "$download_base$dropbox_folder/$filename";
+    my $sip_directory = sprintf "%s/%s", $volume->get_sip_directory(), $volume->get_namespace();
+    if (not -d $sip_directory) {
+      get_logger->trace("Creating download directory $sip_directory");
+      mkdir($sip_directory, 0770) or $self->set_error('OperationFailed', operation=>'mkdir', detail=>"$sip_directory could not be created");
+    }
+    my $cmd = $self->rclone_command($rclone_config_path, $url, $sip_directory);
+    get_logger->trace("Running $cmd");
+    my $output = `$cmd `;
+    if (${^CHILD_ERROR_NATIVE} != 0) {
+      $self->set_error('StageFailed', detail => $output);
+      return 0;
+    }
+    my $outcome = PREMIS::Outcome->new('pass');
+    $volume->record_premis_event("package_inspection", outcome => $outcome);
+  }
+  $self->_set_done();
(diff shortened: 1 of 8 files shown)


lib/HTFeed/PackageType/Simple/Download.pm
```diff
@@ -0,0 +1,73 @@
...
+    get_logger->trace("$sip_loc already exists");
+  } else {
+    my $rclone_config_path = get_config('rclone_config_path');
+    unless (defined $rclone_config_path) {
+      $self->set_error('StageFailed', detail => "rclone_config_path not configured");
+      return 0;
+    }
+    my $nspkg = $volume->get_nspkg();
+    my $dropbox_folder;
+    eval {
```
aelkiss: hmm, there should be a way to check this without throwing an exception if it isn't there. Maybe we just set it to `undef` by default in `HTFeed::Namespace`?

lib/HTFeed/PackageType/Simple/Download.pm
```diff
@@ -0,0 +1,73 @@
...
+sub run {
+  my $self = shift;
+
+  my $volume = $self->{volume};
+  # check if the item already exists at $volume->get_sip_location()
+  my $sip_loc = $volume->get_sip_location();
+  if (-f $sip_loc) {
+    get_logger->trace("$sip_loc already exists");
+  } else {
+    my $rclone_config_path = get_config('rclone_config_path');
```
aelkiss: Probably worth extracting the contents of the else clause to a separate method, esp. since it isn't really feasible to have the `if (-f $sip_loc)` as a `return if ... ` -style guard clause (since we still have to do the stuff at the end)

lib/HTFeed/PackageType/Simple/Download.pm
```diff
@@ -0,0 +1,73 @@
...
+      mkdir($sip_directory, 0770) or $self->set_error('OperationFailed', operation=>'mkdir', detail=>"$sip_directory could not be created");
+    }
+    my $cmd = $self->rclone_command($rclone_config_path, $url, $sip_directory);
+    get_logger->trace("Running $cmd");
+    my $output = `$cmd `;
+    if (${^CHILD_ERROR_NATIVE} != 0) {
+      $self->set_error('StageFailed', detail => $output);
+      return 0;
+    }
+    my $outcome = PREMIS::Outcome->new('pass');
```
aelkiss: I think this is copy-pasta from the IA download stage and not relevant here perhaps -- may be worth reviewing the Google download routines https://github.com/hathitrust/feed_google/blob/master/lib/HTFeed/PackageType/Google/Download.pm

aelkiss (changes requested):
Looks pretty good; just a couple of things to clean up.

Regarding CI failing I think we just need to add `dropbox_folder` to https://github.com/hathitrust/feed/blob/master/etc/sample_namespace/TEST.pm -- I don't think the tests currently run using all the private submodules.

lib/HTFeed/PackageType/Simple/Download.pm
moseshll: Is stage failure nonetheless the expected behavior if there is no dropbox folder configured for the namespace?

lib/HTFeed/PackageType/Simple/Download.pm
moseshll: Oh heck yes :-)
(But I did scratch my head and wonder if I should keep the PREMIS code intact since IIRC the virus scan stage I did for EMMA also did something not dissimilar)

lib/HTFeed/PackageType/Simple/Download.pm
aelkiss: If the file doesn't exist and the dropbox folder isn't configured, then yes, it should fail with a message to that effect.

moseshll: The only "clean" way I see to modify `HTFeed::Namespace` is to add a guard clause in the `get` inner exception handler (https://github.com/hathitrust/feed/blob/master/lib/HTFeed/Namespace.pm#L76), since the logic there is using `if (defined ...)` on everything: initializing `dropbox_folder => undef` will just fall through to a `croak` there anyway. So I think I'm gonna keep the `eval` bit unless there's an avenue I haven't explored.

(add chin-rubbing emoji here...)

Maybe adding a non-exception-raising "has_key" method akin to Ruby `key?` in `HTFeed::Namespace` would restore some order to the insanity?

aelkiss: Hmmm.. Maybe what we want is `exists` instead of `defined` so we don't end up throwing an exception when something is present but `undef`; I very much doubt we're relying anywhere on the existing behavior. I'll push a commit with the proposed change.

aelkiss: I pushed a commit with some cleanup here; tests should pass.

lib/HTFeed/PackageType/Simple/Download.pm
aelkiss: Took care of this in the commit I pushed as well.

aelkiss (approved):
Looks good to me once tests pass.

PR #72, opened 2021-10-04
HT-3097 Autoremove material from Dropbox after ingest

moseshll:
- Move common rclone-related operations into `HTFeed::Rclone`
- Move `rclone_url` into the Simple package `Volume` class
- Add `delete` subcommand and a bit of sanity checking to `t/bin/rclone_stub.pl`

### DIFF
--- a/t/local_ingest.t
+++ b/t/local_ingest.t
@@ -124,18 +124,16 @@ describe "HTFeed::PackageType::Simple::Download" => sub {
     $testlog = HTFeed::Test::Logger->new();
     set_config(0,'stop_on_error');
     set_config($tmpdirs->test_home . "/fixtures/rclone_config.conf", 'rclone_config_path');
+    set_config("$FindBin::Bin/bin/rclone_stub.pl", 'rclone');
   };
 
   before each => sub {
     $tmpdirs->setup_example;
     $testlog->reset;
-    $save_rclone = $ENV{rclone};
-    $ENV{rclone} = "$FindBin::Bin/bin/rclone_stub.pl";
   };
 
   after each => sub {
     $tmpdirs->cleanup_example;
-    $ENV{rclone} = $save_rclone;
   };
 
   after all => sub {
@@ -160,7 +158,6 @@ describe "HTFeed::PackageType::Simple::Volume" => sub {
   use HTFeed::PackageType::Simple::Download;
   my $tmpdirs;
   my $testlog;
-  my $save_rclone;
   my $fetchdir;
 
   before all => sub {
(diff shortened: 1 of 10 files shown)


t/local_ingest.t
```diff
@@ -145,11 +150,61 @@ describe "HTFeed::PackageType::Simple::Download" => sub {
...
+  };
+
+  describe "#clean_sip_success" => sub {
+    it "calls rclone to remove SIP from Dropbox" => sub {
+      my $volume = HTFeed::Volume->new(
+        namespace => 'test',
+        objid => 'test_objid',
+        packagetype => 'simple');
+      eval {
+        $volume->clean_sip_success();
```
aelkiss, line 198: The stub doesn't really do anything, so I think this test is just testing that `clean_sip_success` is defined.

It strikes me that now that `HTFeed::RClone` has its own class, we could mock that instead (e.g. using https://metacpan.org/pod/Test::Spec::Mocks) but I'm not sure it's worth the trouble given that we have the rclone stub  binary already. I suppose another option would be to have the rclone stub use its own temp directory for tests, and then it could actually implement delete.

Both options seem a little bit painful for what we'd gain given that we're really only testing one line of code (e.g. ensuring that rclone is called the way we expect) and that we aren't getting an integration test here, so I might have  talked myself into not worrying about testing this. I'm not sure where that leaves this test; I would be inclined just to remove it.

aelkiss: I guess if we did inject a mock object into the stage / volume for use in the tests, we could use that for testing both the copy & the delete, which might be worthwhile; I think that would probably be preferable to continuing to build out `bin/rclone_stub.pl`. We could then perhaps test `RClone.pm` itself by injecting the command to run as `echo` or something.

aelkiss (commented):
I like separating out the rclone interface to `HTFeed::RClone`. You could consider adding `copy` / `delete` methods there. I'm waffling on how/whether we should do additional testing of the interaction with rclone.

t/local_ingest.t
moseshll: I like having the external binary because it throws a different codebase at sanity-checking the parameters, even though it's otherwise more or less of a no-op, and I think in that way it's more potent than an `echo`

moseshll: I would prefer to have the rclone binary set via config rather than environment variable, I think. (And it may make sense to do so given we have some configs like Jhove and Kakadu that have/need directions to find the executable. But may be overkill to require it.)

aelkiss: Sounds reasonable. There's still the question of whether/how to test that `clean_sip_success` actually calls delete.

moseshll: I can at least check the log for a signal that the binary is called with a `delete` subcommand.

moseshll:
Some further refinements for the branch.

aelkiss:
Works for me. I'll update the namespaces submodule & test this out in dev with something in the /ULIB-hathitrust-ingest/test-hathitrust-ingest dropbox folder.

aelkiss:
A bunch of yak shaving later (with both rclone config & getting everything set up in the database tables & fixtures so that I can run through the whole ingest process for a test item), I've verified that both the download & the deletion from Dropbox works. I did need to add the actual rclone binary to the docker image.

aelkiss (approved):

PR #73, opened 2021-10-19
HT-3209 Backup expiration logic is wrong

moseshll:
- First SQL query (clustering) makes sure there are both new and old versions.
- Second SQL query (deletion candidates) uses an offset=1 to preserve the most recent old version.
- Tests updated.

### DIFF
--- a/lib/HTFeed/BackupExpiration.pm
+++ b/lib/HTFeed/BackupExpiration.pm
@@ -31,26 +31,20 @@ sub run {
 
   my $config = get_config('storage_classes');
   my $storage_config = $config->{$self->{storage_name}};
-  # Get a list of all volumes with this storage name that have at least one new version
-  # and at least one old version, and at least three in all. Those that have multiple old
-  # versions will be culled below.
+  # find everything with more than one version that is at least 6 months old
+  # delete all but the most recent > 6 months old version
   my $sth = get_dbh()->prepare(<<'SQL');
     SELECT namespace,id
     FROM feed_backups
     WHERE deleted IS NULL
       AND storage_name=?
+      AND version < DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
     GROUP BY namespace,id
-    HAVING COUNT(*) > 2
-      AND MIN(version) < DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
-      AND MAX(version) >= DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
+    HAVING COUNT(*) > 1
 SQL
 
   $sth->execute($self->{storage_name});
   while (my $row = $sth->fetchrow_hashref) {
-    # Get the versions in this group that can be deleted: select only the old versions,
-    # sort from most recent and offset by one to exclude the newest.
-    # MySQL reference gives an example of using an arbitrary large number to
-    # fetch remainder after offset, seems pretty sketchy to me.
     my $versions_sth = get_dbh()->prepare(<<'SQL');
       SELECT version
       FROM feed_backups
@@ -59,16 +53,18 @@ SQL
         AND namespace=?
         AND id=?
         AND version < DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
-        ORDER BY version DESC LIMIT 1,999999999
+        ORDER BY version DESC
 SQL
     $versions_sth->execute($self->{storage_name}, $row->{namespace}, $row->{id});
-    foreach my $version (map { $_->[0]; } @{$versions_sth->fetchall_arrayref}) {
+    my @versions = map { $_->[0]; } @{$versions_sth->fetchall_arrayref};
+    shift @versions; # jettison the most recent
+    foreach my $version (@versions) {
       my $volume = new HTFeed::Volume(namespace => $row->{namespace},
                                       objid => $row->{id},
                                       package_type => 'ht');
-      my $storage = $storage_config->{class}->new(volume    => $volume,
-                                                  config    => $storage_config,
-                                                  name      => $self->{storage_name});
+      my $storage = $storage_config->{class}->new(volume => $volume,
+                                                  config => $storage_config,
+                                                  name   => $self->{storage_name});
       unless (defined $storage) {
         die "Unable to get storage for $volume->{namespace}.$volume->{objid}";
       }
--- a/lib/HTFeed/BackupExpiration.pm
+++ b/lib/HTFeed/BackupExpiration.pm
@@ -31,22 +31,26 @@ sub run {
 
   my $config = get_config('storage_classes');
   my $storage_config = $config->{$self->{storage_name}};
-  # Get a list of all volumes with this storage name that have one or more old
-  # versions. This query is about finding sets of volumes, nothing more granular.
+  # Get a list of all volumes with this storage name that have at least one new version
+  # and at least one old version, and at least three in all. Those that have multiple old
+  # versions will be culled below.
   my $sth = get_dbh()->prepare(<<'SQL');
-    SELECT namespace,id,MAX(version) AS max_version
+    SELECT namespace,id
     FROM feed_backups
     WHERE deleted IS NULL
       AND storage_name=?
     GROUP BY namespace,id
-    HAVING COUNT(*) > 1
+    HAVING COUNT(*) > 2
       AND MIN(version) < DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
+      AND MAX(version) >= DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
 SQL
 
   $sth->execute($self->{storage_name});
   while (my $row = $sth->fetchrow_hashref) {
-    # Get the versions in this group that can be deleted:
-    # Everything except the newest (regardless of its age), and except new versions.
+    # Get the versions in this group that can be deleted: select only the old versions,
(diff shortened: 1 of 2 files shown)


aelkiss:
I'm having a hard time following the logic at a first glance and need to reason through it a bit. I might want to make the tests somewhat more concrete; not sure at this point.

aelkiss:
Or maybe I just want to change the description of the tests a little bit..

t/backup_expiration.t
```diff
@@ -119,38 +119,49 @@ describe "HTFeed::BackupExpiration" => sub {
       is($exp->{storage_name}, $vars{storage_name}, 'expiration has correct storage name');
     };
 
-    it "does not do anything with a single version" => sub {
+    it "does not do anything with a single old version" => sub {
```
aelkiss: "when the only version was backed up > 6 months ago, does not delete it"

t/backup_expiration.t
```diff
@@ -119,38 +119,49 @@ describe "HTFeed::BackupExpiration" => sub {
...
       my $deleted = count_deleted_objects('test', 'test', $storage->{timestamp});
       is($deleted, 0, 'object is not deleted');
-      ok(!mets_deleted($storage), "new ($storage->{timestamp}) mets left intact");
-      ok(!zip_deleted($storage), "new ($storage->{timestamp}) zip left intact");
+      ok(!mets_deleted($storage), "single ($storage->{timestamp}) mets left intact");
+      ok(!zip_deleted($storage), "single ($storage->{timestamp}) zip left intact");
     };
 
-    it "deletes old versions when there is a new one" => sub {
+    it "does not do anything with a single old version and single new one" => sub {
```
aelkiss: "when the newest version was backed up < 6 months ago, does not delete the prior version"

t/backup_expiration.t
```diff
@@ -119,38 +119,49 @@ describe "HTFeed::BackupExpiration" => sub {
...
+      ok(!mets_deleted($old_storage), "old ($old_storage->{timestamp}) mets left intact");
+      ok(!zip_deleted($old_storage), "old ($old_storage->{timestamp}) zip left intact");
+
+      $deleted = count_deleted_objects('test', 'test', $new_storage->{timestamp});
+      is($deleted, 0, 'new object is not deleted');
+      ok(!mets_deleted($new_storage), "new ($new_storage->{timestamp}) mets left intact");
+      ok(!zip_deleted($new_storage), "new ($new_storage->{timestamp}) zip left intact");
+    };
+
+    it "does not delete old versions when there is nothing new" => sub {
```
aelkiss: I'm not sure this is correct. Assume today is 2021-10-21 and consider this case:

version A - deposited 2021-04-19
version B - deposited 2021-04-20

In this case, once version B is deposited, we need to retain version A until 2021-10-20 (i.e. we need to retain A for 6 months after B is backed up)

Assume we don't run this cleanup operation until 2021-10-21. In that case, both versions will be considered "old" (since they are > 6 months old); there is no "new" (< 6 month old version), but we no longer need to retain version A in this case. I think the logic here is reliant on us actually running this frequently enough to catch it in the state where version B is "new" and version A is "old", which would only happen if we ran this at the right time on 2021-10-20.

t/backup_expiration.t
```diff
@@ -119,38 +119,49 @@ describe "HTFeed::BackupExpiration" => sub {
...
+        is($deleted, 0, 'old object is not marked feed_backups.deleted');
+        ok(!mets_deleted($old_storage), "old ($old_storage->{timestamp}) mets left intact");
+        ok(!zip_deleted($old_storage), "old ($old_storage->{timestamp}) zip left intact");
       }
-      ok(!mets_deleted($storage), "new ($storage->{timestamp}) mets left intact");
-      ok(!zip_deleted($storage), "new ($storage->{timestamp}) zip left intact");
     };
 
-    it "keeps the newest old version" => sub {
+    it "deletes the oldest old version when there is a new one" => sub {
```
aelkiss: Again, I don't think this implements the desired logic in all cases; deletion shouldn't be dependent on having a < 6 month old version, just on there being multiple > 6 month old versions.

aelkiss (changes requested):
Let's go back to the criteria in the ticket:

"retain old versions for 6 months after a newer version has been backed up"

I still think the logic as described in the ticket is correct:

* find everything with more than one version that is at least 6 months old
* delete all but the most recent > 6 months old version

but I don't think that's quite what's implemented here. I'll add a couple additional notes inline.

lib/HTFeed/BackupExpiration.pm
```diff
@@ -31,22 +31,26 @@ sub run {
...
   my $config = get_config('storage_classes');
   my $storage_config = $config->{$self->{storage_name}};
-  # Get a list of all volumes with this storage name that have one or more old
-  # versions. This query is about finding sets of volumes, nothing more granular.
+  # Get a list of all volumes with this storage name that have at least one new version
+  # and at least one old version, and at least three in all. Those that have multiple old
+  # versions will be culled below.
   my $sth = get_dbh()->prepare(<<'SQL');
-    SELECT namespace,id,MAX(version) AS max_version
+    SELECT namespace,id
```
aelkiss, line 37: I think we just need to find everything with at least two versions that are more than 6 months old, regardless of any versions < 6 months old (which we never touch):

```sql
SELECT namespace, id
FROM feed_backups
WHERE deleted IS NULL
    AND storage_name=?
    AND version < DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
    GROUP BY namespace,id
    HAVING count(*) >= 2
```

lib/HTFeed/BackupExpiration.pm
```diff
@@ -55,10 +59,9 @@ SQL
         AND namespace=?
         AND id=?
         AND version < DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
-        AND version!=?
+        ORDER BY version DESC LIMIT 1,999999999
```
aelkiss: This works, but I agree we don't really want the magical number here... Could we instead just `ORDER BY version DESC` and skip the first row of the results on the application side?

t/backup_expiration.t
aelkiss: consider as a description "deletes old versions after 6 months after a newer version has been backed up"

t/backup_expiration.t
```diff
@@ -160,18 +171,19 @@ describe "HTFeed::BackupExpiration" => sub {
       if ($old_versions[0]->{timestamp} > $old_versions[1]->{timestamp}) {
         ($newer, $older) = @old_versions;
       }
+      my $storage = prepare_storage($vars{storage_name}, new_random_timestamp());
```
aelkiss: Shouldn't require a version < 6 months old -- perhaps consider in this case having 3 versions > 6 months old and ensuring all but the most recent of those gets deleted

aelkiss:
As far as cases in the tests, we probably want to make sure these are covered:

- multiple < 6 month old versions --> none deleted
- single > 6 month version  and single < 6 month old version --> none deleted
- multiple > 6 month old versions --> retains the most recent version & the other versions are deleted
- multiple > 6 month old versions & < 6 month old versions --> retains the most recent > 6 month old version and all < 6 month old versions & deletes the others

t/backup_expiration.t
moseshll: I'm keeping separate tests with and without "new" versions since that's been the crux of this latest round of confusion. Using the descriptions "deletes the oldest old versions when there are no new ones" and "deletes the oldest old versions when there are new ones". Since I seem to have standardized on "old" to mean > 6mo and "new" to mean <6mo... and where needed "newer" to substitute for "newest old".

t/backup_expiration.t
moseshll: Trying to open up descriptions with a verb coming after the Test::Spec "it" operator. So for this one I am trying out "deletes nothing with a single old version and single new one"

lib/HTFeed/BackupExpiration.pm
```diff
@@ -59,16 +53,18 @@ SQL
         AND namespace=?
         AND id=?
         AND version < DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
-        ORDER BY version DESC LIMIT 1,999999999
+        ORDER BY version DESC
 SQL
     $versions_sth->execute($self->{storage_name}, $row->{namespace}, $row->{id});
-    foreach my $version (map { $_->[0]; } @{$versions_sth->fetchall_arrayref}) {
+    my @versions = map { $_->[0]; } @{$versions_sth->fetchall_arrayref};
+    shift @versions; # jettison the most recent
```
aelkiss, line 60: We might want to clarify the comment here -- "jettison" isn't perhaps the best word since we are in fact *retaining* that version.. perhaps:

```perl
shift @versions; # ensure we don't delete the most recent
```

aelkiss (commented):
I think this looks correct now, but I want to take one more look through when it's not the end of the day...

t/backup_expiration.t
```diff
@@ -119,74 +119,87 @@ describe "HTFeed::BackupExpiration" => sub {
...
-      foreach my $n (1 .. 2) {
+      foreach my $n (1 .. 3) {
         my $storage = prepare_storage($vars{storage_name}, old_random_timestamp());
         push @old_versions, $storage;
       }
-      my ($older, $newer) = @old_versions;
-      if ($old_versions[0]->{timestamp} > $old_versions[1]->{timestamp}) {
-        ($newer, $older) = @old_versions;
-      }
+      @old_versions = sort { $a->{timestamp} cmp $b->{timestamp}; } @old_versions;
```
aelkiss, line 157: This looks correct; I wonder though if it would be clearer if we explicitly specified the timestamps for each object rather than including this logic here.

aelkiss (approved):
I took another look and I'm satisfied this matches the criteria (when a new version is deposited, the previous version is retained for 6 months). I will go ahead and merge it. Thanks!

PR #75, opened 2021-10-27
Make a node-specific staging directory

aelkiss:
This is helpful if we're using a shared filesystem for temporary space,
and doesn't harm anything if we aren't.


aelkiss:
FYI @botimer

PR #77, opened 2021-11-09
HT-3217 Reading order for IA objects

moseshll:
I created a new file (t/ia.t) for the tests because I was at a loss on how to graft the new fixtures onto the t/lib/HTFeed/PackageType/IA family.


aelkiss:
I think that's exactly right regarding a new test file.

aelkiss (approved):
Looks exactly right. I'll give it a try in development and (assuming it looks good) go ahead and merge it.

aelkiss:
OK, managed to run this on the real object in dev, and it looks like the reading order is in the source METS but doesn't get copied to the HT METS. I'll take a look and see why that is.

aelkiss:
Also it looks like there's some inconsistency between using `dmdSec` and `techMD`. I'll see what makes sense.

aelkiss:
@moseshll  could you take a look at the commit I pushed?

Basically, I found that:

- reading order was extracted from source METS on a package-type by package-type basis, not in a consistent way

- in part that was because there was some handling for old variants of Google-provided reading order metadata that we shouldn't be getting any more, which I excised

- I also noticed reading order was in a dmdSec rather than a techMD in the source METS for the "simple" package type, but it wasn't clear why -- it's in a `techMD` in the Google METS, and that's where we were putting it in the HathiTrust METS as well. Now it is consistently in techMD.

moseshll:
Runs fine here. Merging.

PR #79, opened 2021-11-29
HT-3183: Add dry-run flag for backup expiration

aelkiss:
This is so we can verify what it would delete in production without actually deleting anything.

### DIFF
--- a/lib/HTFeed/BackupExpiration.pm
+++ b/lib/HTFeed/BackupExpiration.pm
@@ -10,6 +10,9 @@ use Carp;
 use Log::Log4perl qw(get_logger);
 use File::Temp;
 
+use HTFeed::Storage::PrefixedVersions;
+use HTFeed::Storage::ObjectStore;
+
 sub new {
   my $class = shift;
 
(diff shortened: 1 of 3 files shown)


aelkiss:
Also fixed a couple of includes. I don't love sprinkling the concrete storage class names as requires in multiple places; maybe it makes sense just to require them from `HTFeed::Storage`?

lib/HTFeed/BackupExpiration.pm
```diff
@@ -43,18 +50,27 @@ sub run {
     HAVING COUNT(*) > 1
 SQL
 
+  my $versions_sth = get_dbh()->prepare(<<'SQL');
```
aelkiss, line 55: Small optimization, but I noticed we were re-preparing the statements for every object, and it didn't seem like we needed to do that. @moseshll I did want to check there wasn't any particular reason for that though.

moseshll (approved):
The only thing that I find questionable is something I failed to implement: a sanity check on  `defined $storage_config` early in the process, maybe as part of the constructor. As far as what's actually here in this PR, runs fine for me. (I should elaborate: if you pass a bogus storage name on the CLI the script fails silently.)

lib/HTFeed/BackupExpiration.pm
moseshll: No, my experience with the MySQL DBI in Perl has been to sidestep statement handles, so I don't tend to use them in particularly elegant ways.

aelkiss:
Thanks for the quick turnaround. I'll add the check for `defined $storage_config`.

PR #80, opened 2021-12-01
Support for new IA minter

aelkiss:
Also adds support for new NRLF barcode scheme


aelkiss:
FYI @jsjiang @cscollett

PR #81, opened 2021-12-07
HT-3210: progress tracker via prometheus / pushgateway

aelkiss:
- add ProgressTracker for use by cron jobs
- add pushgateway for docker-compose for development
- update google submodule to return metrics

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ RUN apt-get update && apt-get install -y \
     awscli \
     epubcheck \
     libclamav-client-perl \
+    libnet-prometheus-perl \
     libswitch-perl \
     libtest-class-perl \
     libtest-mockobject-perl \
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ RUN apt-get update && apt-get install -y \
     libtest-class-perl \
     libtest-mockobject-perl \
     libtest-most-perl \
+    libtest-time-perl \
     libtest-spec-perl \
     mariadb-client \
     netcat \
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -9,10 +9,11 @@ services:
       - FEED_HOME=/usr/local/feed
       - TEST=1
       - VERSION=feed-testing
-    command: bash -c "bin/wait-for --timeout=300 mariadb:3306 minio:9000 -- prove -I lib"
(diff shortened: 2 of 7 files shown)


aelkiss:
FYI @daaang , @jsteverman 

Compare to https://github.com/hathitrust/holdings-backend/blob/main/lib/utils/push_metrics_waypoint.rb (which doesn't yet implement the 'stage' idea)

@moseshll  Do you think it's worth adding tests for this, and if so, what's your sense of what the best option for testing would be? For holdings, we didn't add unit tests for PushMetricsWaypoint (above), but some of the existing tests at least exercise the code, and we stub the request to prometheus using webmock. If we do test this, I'm not sure what the best option for doing that is - the most thorough would be to run the pushgateway image in docker, push actual metrics to it, and then test that they show up if we curl the pushgateway. A pure unit testing approach where we try to stub `Net::Prometheus` and individual metrics seems kind of fiddly. A middle ground might be to stub the LWP useragent and test that we're posting what we expect.

moseshll:
If the pushgateway docker addition here is configured and working, I'd say use that as it's the closest fit between dev/test and IRL. I understand that's a big "if" at this point.

aelkiss:
:+1: Yes, it is working (there isn't really any configuration needed for it.) I'll see if I can add a test for that.

aelkiss:
Added tests.

In this case all the tests just look at the exposed metrics from the pushgateway.

We could probably change some of the tests to look at what gets pushed to the pushgateway (i.e. the output of `$self->{prom}->render`), but these tests are 1) fast enough as-is, 2) consistent, and 3) avoids confusion between what the internal state is vs. what's actually been pushed to the pushgateway.

@moseshll , @daaang : thoughts?

aelkiss:
Some backup expiration tests are now failing, presumably because of using `Test::Time`. As in Ruby, the dependencies leak between tests... I'll take a look.

aelkiss:
In inspecting I can't replicate the failure; I suspect it has to do with the vagary of randomness in the test setup; perhaps it happened to end up with two versions with the same timestamp. I'm going to try re-running the tests and see if it recurs. If so, there's probably some difference between the CI environment and my local one that I don't understand. If not, this is probably an issue with test setup, and we could perhaps consider either reducing the randomness in test setup or at least logging what that randomness was so we can reproduce.

moseshll:
I also got the test failures and didn't pay much mind yesterday since it was a WIP. I'll have a look after lunch.

moseshll (approved):
Looks good. No errors here.

PR #82, opened 2022-01-03
Tests for private submodules

aelkiss:
Object ID validity is pretty annoying because it requires a private
submodule to test, but the test can't really live in the submodule
because the submodule only has the actual namespace configuration, not
the whole ingest module. It might be worth thinking about some
alternatives to the private submodule model, but I'm not sure what they
would be.

In this case, there were some uncaught errors in the namespace
configuration. This reintroduces and updates some tests for object ID
validity and adds a hook for tests that require the private
submodules to run.


aelkiss:
FYI @moseshll. I'm going to go ahead and merge this assuming tests pass, because stuff is broken for ingest & audit for items in the uc2 and ia namespaces right now, but please feel free to comment on the approach and we can adjust as necessary.

PR #83, opened 2022-01-05
Track namespace & app labels for pushgateway

aelkiss:
Adds optional additional labels for job tracking via pushgateway.


moseshll (approved):
Looks good and tests work for me.

PR #84, opened 2022-01-06
RabbitMQ for ingest queue

aelkiss:
- tests job queue functionality for rabbitmq
- starts moving single_thread / runlite stuff to QueueRunner (not
tested)


aelkiss:
FYI @billdueber -- starting to look at using rabbitmq for ingest queue

aelkiss:
I think this is ready for review. Here is a basic overview of what happened before & after:

- Before, there was a table (`feed_queue`) which listed items & their state
- Now, we still use that table for tracking item status, but use the RabbitMQ queue to find items ready for work
- When an item is enqueued (and added to `feed_queue`) we also send a message to RabbitMQ
- Worker threads get items to work on from RabbitMQ, work on them until they're done, update the status in the database, and acknowledge the message from RabbitMQ.
- Tracking success/failure is out of the scope of the queuing implementation -- the `feed_queue` table still tracks success & failure, and (for now) if we want to re-queue failures, that would be the responsibility of a separate process.
- Nothing in particular forces `feed_queue` and RabbitMQ to be in sync
- RabbitMQ will automatically re-deliver unacknowledged messages (i.e. if a worker unexpectedly crashes)

I haven't tried running this end-to-end in development; that would be the next step before trying to deploy this in production.

billdueber:
@aelkiss I'll dig into this, but it's gonna take a while :-)

aelkiss:
Yup, no worries. I'll see if I can get something on our calendars next week to go over this a bit.

aelkiss:
I've tried a couple things:

- queued 5,000 test IA items
- verified this results in 5,000 messages in rabbitmq
- verified that if I ctrl-C a running worker, the status does not get updated in the database, and that job gets re-delivered when I start a new worker

I think I need to go back and test more some of the things related to job progression -- I'm not seeing it progress beyond the download stage for items, but rather do the download, retry download several times (??), ack the message, and then leave things in state 'downloaded', so I need to figure out what's going on there.

lib/HTFeed/Queue.pm
```diff
@@ -0,0 +1,285 @@
...
+
+    unless ($args{no_bibdata_ok} || $self->has_bibdata($volume)) {
+      $self->{dbh}->rollback && return 0 
+    }
+
+    if ($args{use_disallow_list} && $self->disallowed($volume)) {
+      $self->{dbh}->rollback && return 0;
+    }
+
+    my $priority = initial_priority($volume,$priority_modifier);
```
aelkiss: Should remove references to priority since RabbitMQ cannot use it right now

lib/HTFeed/Bunnies.pm
```diff
@@ -0,0 +1,116 @@
...
+  return if $self->{consumer_started};
+  $self->{consumer_started} = 1;
+  # only fetch one item at at time from the queue
+  get_logger()->debug("$BUNNY: started consumer");
+  $self->{mq}->basic_qos($self->{channel},{ prefetch_count => 1 });
+  # enable calling recv; require explicit acknowledgement of message
+  $self->{mq}->consume($self->{channel},$self->{queue},{ no_ack => 0 });
+}
+
+sub next_job {
```
aelkiss, line 81: Should have an option to check that it hasn't been re-delivered too many times

lib/HTFeed/QueueRunner.pm
```diff
@@ -0,0 +1,149 @@
...
+
+sub run {
+  my $self = shift;
+
+  $self->setup_signal_handlers;
+
+  print("feedd QueueRunner running, waiting for something to ingest, pid = $$\n");
+
+  while( my $job_info = $self->{queue}->next_job($self->{timeout}) ){
+    $self->marshal_and_run($job_info);
```
aelkiss, line 57: Should check that the job hasn't been re-delivered too many times already as that might indicate a state where it's just going to crash & fail again.

aelkiss:
Need to make sure that the confirmation for the ack timeout is set longer than the default 30 minutes.

aelkiss:
The installation for `Net::AMQP::RabbitMQ` is mysteriously failing in the build, but @moseshll  otherwise I think this is ready for another look..

Main areas to focus on here is the error handling:

- job isn't decodable:
here https://github.com/hathitrust/feed/pull/84/files#diff-03055aa8a8bd7816cc08fe830772c74256f7d3661824516c8ed36cd36c199574R105 with tests here: 
https://github.com/hathitrust/feed/pull/84/files#diff-a7c4e33aa5a69f961e42f7dcf2594b6c6ef9a5e5d319d4fc51882d1566dcfb65R67-R87

- job isn't something we can handle (bad namespace, id, package type, or stage, or if any of that is missing):
here https://github.com/hathitrust/feed/pull/84/files#diff-f08621d49d82af1f8fe37068b3f5b369e120b4a8602fdcbc60f47c225cee3ba2R75
with tests here https://github.com/hathitrust/feed/pull/84/files#diff-f9a0e5be6070e6a8b21f471d5fb24b913dc982f6523b9f7190b81a6f43ef6168R134-R181 

Let me know if it would be helpful to squash this down some before you take another look or if this is OK as-is.

moseshll:
I'm having the same issue testing this as did the automated test. Seems to be an expired certificate issue, if https://github.com/net-amqp-rabbitmq/net-amqp-rabbitmq/issues/220 is to be believed. The timing is certainly right.

aelkiss:
Now the perl module installs, but there's a failing test in GHA I can't reproduce locally...

```
#   Failed test 'HTFeed::Bunnies #next_job saves a reference to the message'
#   at t/bunnies.t line 64.
```

I'll try re-running the tests in GHA and see if it's reproducible there or not. If it is, there's a way to get a shell in the GHA runner now, which might be helpful to see in more detail what's going wrong.

aelkiss:
Succeeded on re-try, so I'm inclined to chalk this up to cosmic rays or whatever. If we see it again in GHA or locally we should certainly try to debug the issue.

aelkiss:
@moseshll Do you want to give this another try? If it looks OK & you're happy with the error handling, I'll go ahead and squash & merge this.

moseshll:
@aelkiss yes, I'd like to have a go at running on the work laptop this afternoon, then give the code another once-over.

moseshll (approved):
Still can't meaningfully test this on my arm64 work machine (like an idiot I deleted the repo from my personal laptop) because perl_base and db-image/mariadb are running under emulation. Not sure which is the bottleneck at this point.

aelkiss:
I will work on rebasing, merging, and deploying this tomorrow.

aelkiss:
@moseshll  rebased, squash down a bit; will merge once tests pass

PR #86, opened 2022-02-24
HT-3330: Support for testing FTPS in development

aelkiss:
- curl
- ftps server in docker-compose
- config for using ftps server

The curlrc automatically adds --insecure (-k) in development since we
only have a self-signed certificate.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ ingest_stage/*
 ingest_sips/*
 .gnupg
 .bash_history
+var/*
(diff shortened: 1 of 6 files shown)


aelkiss:
FYI @niquerio -- maybe of interest for testing sending holdings to zephir

aelkiss:
See also https://github.com/hathitrust/feed_jobs/pull/18

aelkiss:
Looks like Zephir actually uses vsftpd, so I'll see if I can find a docker image for that & get that working with docker-compose with FTPS

aelkiss:
I poked a bit at vsftpd, but I didn't find anything that seemed particularly well-maintained; I tried generating my own vsftpd config but ran into a bunch of yak-shaving. I think this is good enough for now, and if we find some incompatibility in the future we can do something different.

jsteverman (approved):

PR #89, opened 2022-03-07
HT-3375: run clamscan on demand

aelkiss:
Given the low number of files we need to scan (only items coming from
EMMA), running clamav in a sidecar container for every ingest worker is
a waste of resources; especially, downloading virus definitions for
every separate pod. This assumes virus definitions will be present at
/var/lib/clamav and calls out to clamscan when needed. It preserves the
same interface as ClamAV::Client, so switching back should be easy if
needed in the future.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@ LABEL org.opencontainers.image.source https://github.com/hathitrust/feed
 
 RUN apt-get update && apt-get install -y \
     awscli \
+    clamav \
     curl \
     epubcheck \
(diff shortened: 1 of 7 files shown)


aelkiss:
This will need to be accompanied by a change here: https://github.com/hathitrust/ht_tanka/blob/main/lib/ingest/workers.libsonnet to mount a directory from NFS with virus definitions to `/var/lib/clamav` as well as something to keep those up-to-date.

moseshll (approved):
Beautifully simple.

PR #94, opened 2022-04-07
Remove submodules

aelkiss:
First draft of cleaning up dockerfile & submodule situation for ingest. There will be another new repository for the stuff that was in the 'jobs', 'google', and 'namespace' submodules (those dependencies really were going the wrong direction') that pulls in this repository as a module in the same way this one pulls in metslib. I'm hopeful we might be able to use something like this pattern for mdp-lib in the future.


aelkiss:
Still to-do:

- get grok via hathitrust.github.io
- test that we can compress & decompress images with it

.github/workflows/build-manual.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     - name: Check that the tag exists in container registry
       id: image_check
       run: |
-        if docker manifest inspect ghcr.io/hathitrust/feed-unstable:${{ steps.tag_check.outputs.tag }} > /dev/null; then
+        if docker manifest inspect ghcr.io/hathitrust/ingest-unstable:${{ steps.tag_check.outputs.tag }} > /dev/null; then
```
aelkiss: need to revert this back to feed-unstable - false path down changing the name

README.md
```diff
@@ -1,12 +1,11 @@
-HathiTrust Ingest Toolkit (Feed)
+HathiTrust Ingest Toolkit
 
-![Run CI](https://github.com/hathitrust/feed/actions/workflows/ci.yml/badge.svg)
-![Docker Build](https://github.com/hathitrust/feed/actions/workflows/build.yml/badge.svg)
+![Run CI](https://github.com/hathitrust/ingest/actions/workflows/ci.yml/badge.svg)
```
aelkiss: need to fix this here too. :weary:

bin/compress_tif_jp2.pl
```diff
@@ -43,17 +43,13 @@
 
     # try to compress the TIFF -> JPEG2000
     print("Compressing $infile to $outfile\n");
-    my $kdu_compress = get_config('kdu_compress');
-    die("You must correctly configure the path to kdu_compress in the feed configuration\n") 
-    if not defined $kdu_compress or !-x $kdu_compress;
+    my $grk_compress = get_config('grk_compress');
```
aelkiss: change to grok should be a separate PR

lib/HTFeed/FactoryLoader.pm
```diff
@@ -118,56 +124,69 @@ sub import {
...
-        if($@) { die $@ };
-    }
+  # determine the subdirectory to find plugins in
+  my $caller = shift;
+  my $module_path = module_path($caller);
+  my $relative_path = relative_path($caller);
+
+  load_all_subclasses_from_dir($caller,$module_path,$relative_path);
 
-#        warn("Factory loading in $module_path\n");
```
aelkiss: revert whitespace change

README.md
```diff
@@ -1,12 +1,11 @@
...
+HathiTrust Ingest Toolkit
 
 ![Run CI](https://github.com/hathitrust/feed/actions/workflows/ci.yml/badge.svg)
 ![Docker Build](https://github.com/hathitrust/feed/actions/workflows/build.yml/badge.svg)
 
 ```bash
 git clone https://github.com/hathitrust/feed
-cd feed
-git submodule update --init metslib
+cd ingest
```
aelkiss, line 8: wrong

aelkiss:
I think the most relevant things to note are:

- removing the 'base' image and specifying the dependencies we actually need in the Dockerfile and in Makefile.PL
- pulling in the METS stuff via `cpanm https://github.com/hathitrust/metslib.git@v1.0.1` - https://github.com/hathitrust/feed/pull/94/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R77
- using shared github actions via github.com/hathitrust/github_actions

aelkiss:
@respinos this is more as an FYI and by way of example for a path that might work for imgsrv/mdp-lib. @moseshll  interested to see if everything works for you; in particular, I haven't built a linux-arm64 version of grok but we could certainly try doing that..

aelkiss:
I tried this out in development & production with this branch and the corresponding work on https://github.com/hathitrust/feed_internal; it appears to be working without error, so I'm going ahead with it.

PR #95, opened 2022-04-11
DEV-10: replace kakadu with grok

aelkiss:
- Uses grok for compressing incoming contone TIFFs to JPEG2000 and for recompressing lossless JPEG2000 images
- Remove unused `kdu_munge` code (see comments)
- Tests compression from contone TIFF and re-compression of lossless JPEG2000


aelkiss:
This produces valid images; still to-do is to pick a PSNR that seems reasonable for our images (requires testing on sample images)

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -557,30 +556,7 @@ sub _remediate_jpeg2000 {
...
-            file      => $outfile,
-            operation => "kdu_munge"
-          );
-
-        $self->update_tags( $exifTool, $outfile );
-    }
-    else {
-        $self->update_tags( $exifTool, $outfile, $infile );
-    }
+    $self->update_tags( $exifTool, $outfile, $infile );
```
respinos, line 559: Does grk_compress not copy over the XMP metadata? Or does it avoid the "large block" problem?

bin/compress_tif_jp2.pl
```diff
@@ -43,17 +43,13 @@
...
+    my $grk_compress = get_config('grk_compress');
+    die("You must correctly configure the path to grk_compress in the feed configuration\n") 
+    if not defined $grk_compress or !-x $grk_compress;
 
-    # Settings for kdu_compress recommended from Roger Espinosa. "-slope"
-    # is a VBR compression mode; the value of 42988 corresponds to pre-6.4
-    # slope of 51180, the current (as of 5/6/2011) recommended setting for
-    # Google digifeeds.
-    system(qq($kdu_compress -quiet -i '$infile' -o '$outfile' Clevels=$levels Clayers=8 Corder=RLCP Cuse_sop=yes Cuse_eph=yes "Cmodes=RESET|RESTART|CAUSAL|ERTERM|SEGMARK" -no_weights -slope 42988))
+    system(qq($grk_compress -i "$infile" -o "$outfile" -p RLCP -n $levels -SOP -EPH -M 62 -I))
```
respinos: So in the end we don't need a slope equivalent for grk_compress? (the slack thread ends on an open question...)

aelkiss: We can provide a target PSNR or compression rate; I'm looking to see what typical PSNR we end up with for scanned book pages with the kakadu slope rate.

lib/HTFeed/Stage/ImageRemediate.pm
aelkiss: exiftool as of 10.03 can handle the "large blocks".

aelkiss: It would probably be worth a test case with a 'huge JPEG2000 box' to ensure we end up with something valid.

t/local_ingest.t
```diff
@@ -110,6 +110,35 @@ describe "HTFeed::PackageType::Simple" => sub {
...
+    };
+
+    it "recompresses lossless jpeg2000 to a valid jpeg2000" => sub {
+      my $volume = unpacked_volume("lossless_jp2");
+
+      HTFeed::PackageType::Simple::ImageRemediate->new(volume => $volume)->run();
+      HTFeed::PackageType::Simple::SourceMETS->new(volume => $volume)->run();
+
+      my $validate = HTFeed::VolumeValidator->new(volume => $volume);
+      eval { $validate->run(); };
```
aelkiss: probably don't need the `eval` here (since we would want the test to fail if `HTFeed::VolumeValidator` raises an exception)

aelkiss:
> This produces valid images; still to-do is to pick a PSNR that seems reasonable for our images (requires testing on sample images)

Done, using requested PSNR 32. This results in similar average PSNR on a variety of scanned book images to kakadu 6.4.0 as evaluated by imagemagick with `compare -metric PSNR`, albeit with about 33% larger total compressed size.

lib/HTFeed/Stage/ImageRemediate.pm
aelkiss: I wasn't able to produce a small test file with a huge JPEG2000 box, but I did manually test with the item you provided & ensure that the output image still contained the original XMP fields. I also added a test around copying over XMP in general; by default, I found that grok doesn't preserve the XMP in all situations, and that `ExifTool::SetNewValuesFromFile` needs an extra option to ensure it preserves the group of the tag (i.e. copies XMP tags to XMP instead of to IFD0 etc)

aelkiss:
This adds a couple of tests around preserving XMP. @respinos  I think this should be good to go after tests are passing, but let me know if anything else comes to mind regarding using grok to create jpeg2000.

PR #96, opened 2022-04-27
DEV-225: Preserve priority from database to rabbitmq

aelkiss:
- Records the priority when queued in the (existing) priority column in
feed_queue

- Passes this priority along to rabbitmq when resetting the item

### DIFF
--- a/bin/enqueue.pl
+++ b/bin/enqueue.pl
@@ -22,7 +22,7 @@ my $insert = 0; # -i
 my $verbose = 0; # -v
 my $quiet = 0; # -q
 # Manually-queued stuff is at high priority by default
-my $priority = 3; # -p
+my $priority = HTFeed::Queue::QUEUE_PRIORITY_HIGH; # -p
 my $state = undef; # -s
 my $help = 0; # -help,-?
 my $use_disallow_list = 1;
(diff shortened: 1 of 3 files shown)


aelkiss:
Note that although nothing really enforces consistency between rabbitmq and the database, this PR doesn't make it any _worse_ -- that is, it was already the case that if something sets an item to `ready` outside of the `HTFeed::Queue::reset` or `HTFeed::Queue::enqueue` methods, a message won't get set to rabbitmq at all. That is still true here, but now we will accept priority and preserve it when an item is reset.

aelkiss:
It would probably be worth adding some constants like `HTFeed::Queue::PRIORITY_HIGH`, etc

aelkiss:
> It would probably be worth adding some constants like `HTFeed::Queue::PRIORITY_HIGH`, etc

Done, force pushed

moseshll (approved):
I could not run tests on ARM64 due to a build failure with grokj2k-tools ("Skipping acquire of configured file 'main/binary-arm64/Packages' as repository 'https://hathitrust.github.io/debian bullseye InRelease' does not seem to provide it (sources.list entry misspelt?)"). So that may not be multi-arch currently.

I'll just go ahead and approve as the sources look quite straightforward and we did go through this code to discuss the finer points.

aelkiss:
Sounds good. I can see about an arm64 build of grokj2k-tools at some future point in time.

PR #98, opened 2022-07-26
DEV-332 CRMS .rights Files Need to be UTF-8

moseshll:
- This patch may be necessary in order to get UTF-8 data into the rights DB.
- However, the added tests pass regardless of the changes to DBTools.pm.
- If the feed_internal populate_rights_data.pl script fails with Unicode text, consider this patch.


aelkiss:
It might be working without the mysql UTF-8 support if essentially it's being treated as binary (no translation) without the flag; not sure. Either way, this should be harmless (if perhaps a bit chicken-wavy)

aelkiss (approved):

PR #100, opened 2022-09-09
Update to JHOVE 1.26.1

aelkiss:
* update JHOVE namespace & xpath queries
* add Devel::Cover
* add ImageMagick settings to support large images
* add coverage & attempt at using coveralls reporter


aelkiss:
@moseshll  I think there are a couple interesting things here, especially in terms of trying to get coverage information - will probably end up merging this before you're back, but perhaps interesting to look at when you have a chance.

aelkiss:
coveralls uploader complains that COVERALLS_REPO_TOKEN is missing. I thought it could use GITHUB_TOKEN but maybe that isn't being set appropriately either.

aelkiss:
Hey, look at that... coverage for perl https://coveralls.io/builds/52404630

@moseshll  we should look into integrating this for post-zephir processing & CRMS as well

PR #101, opened 2022-11-11
DEV-382: Use `ProgressTracker`; other misc fixes

aelkiss:


aelkiss:
FYI @moseshll  regarding using `ProgressTracker` (although you have also seen this on the post-zephir side).

PR #102, opened 2022-12-14
TTO-14 Remove digitizer check by default

aelkiss:
This was relying on hacks in several package types to determine the apparent digitizer, and it's better to check specifically what we can in each package type where it matters.

### DIFF
--- a/bin/ingest_nohandle.pl
+++ b/bin/ingest_nohandle.pl
@@ -60,11 +60,6 @@ if ($fakemeta) {
     return ( 'ht_test','ht_test','ht_test' );
   };
 
-  # don't validate digitizer -- will fail against faked up sources
-  *HTFeed::VolumeValidator::_validate_digitizer = sub {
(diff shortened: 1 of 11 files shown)


aelkiss:
(To be clear: the only package type where we can determine anything about the digitizer with any reliability is Google, and there only Google vs. non-Google. It would be nice if we had a reliable way for material ingested via Internet Archive, since there can be library-uploaded material as well as IA-scanned material, but to date we haven't had a reliable way.)

moseshll (approved):
I also had problems running the tests (collate.t and storage_migrate.t, GitHub action has issues with queue.t) but I don't think my errors are related to these changes.

aelkiss:
OK. Tests passed locally for me. I will try re-running in GHA and see if we get the same results.

aelkiss:
My guess is some race condition/ordering issue with dependent services, but I don't know what specifically.

PR #103, opened 2023-02-28
Prevent TestVolume instances from unnecessarily checking database for…

moseshll:
… validation overrides.


aelkiss (approved):
While I think it might be preferable to remove `note_from_mom` (and maybe some other unused functionality) entirely, this is certainly acceptable for now.

PR #104, opened 2023-04-05
GS-3627 - Decompress before stripping XMP

aelkiss:
There were some images that imagemagick can remediate but that exiftool doesn't want to handle. As far as I can tell, there's no reason to strip the XMP before decompressing, and in fact in testing it seems like ImageMagick strips out the XMP anyway. I can't rule out that there won't be some other image that will fail with this order of operations, but generally this should be harmless.


aelkiss:
FYI @moseshll  - Assuming tests pass, I will merge without review - I'm not sure there's any useful feedback to be had without a deep understanding of all the file formats and the vagaries of all the tools, and the (unanswerable) question is really just what kinds of images will end up tripping up this version of normalization.

PR #105, opened 2023-05-26
GS-5046: JPEG2000 images can have any (reasonable) number of layers

aelkiss:
There isn't any particular reason that we know of to enforce a particular number of layers. We have seen a variety over the years; in the particular instance that inspired this change, there were 3. Rather than enumerating another possible option for something that isn't terribly important, this change just allows any value in the range we've seen.

A consistent number of layers is probably more important for Google-scanned material, where a change in the number of layers could reflect some other change in their image compression process, but I don't think it should matter here for locally-digitized material.


aelkiss:
FYI for maybe @respinos ? Don't think this should affect much of anything.

PR #106, opened 2023-05-26
Don't restore deleted objects

aelkiss:
This works around a race condition where we issue a restore request, but the object has been deleted before the restore request & audit completes successfully.


aelkiss:
FYI @moseshll . Simple enough I think this doesn't really need review if tests pass.

moseshll (approved):
Good to go; this branch fixes an error encountered running against `main`:
```
#   Failed test 'HTFeed::Storage::ObjectStore with encryption enabled stores the mets and encrypted zip'
#   at t/storage_object_store.t line 194.
```

PR #107, opened 2023-09-27
TTO-118: flag to reuse failed item

aelkiss:
Rather than try to encode logic about when it does or doesn't re-download vs. re-use what we already have that varies based on whether the contributor is using dropbox or not, this makes it explicit when re-queuing something whether to use what's already been downloaded or not.

This also removes the logic that (sometimes but only in very specific circumstances/configurations now that we usually check dropbox) tries to automatically re-attempt what's in /htprep/failed, and removes a feature gate that we no longer need (now that all configurations use rabbitmq)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
 volumes_to_test
 ingest_stage
 ingest_sips
+incoming_samples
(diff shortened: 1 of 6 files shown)


aelkiss:
FFR I tested this locally and it worked as expected; we don't currently have automated testing on the queuing script (although it might not be *that* hard to add it.)

PR #110, opened 2024-01-18
Getting feed repo working, shameless green, for mwarin

mwarin:
Mission: to get the `feed` repo to a clean state where @mwarin can start working with it.

Turns out we didn't need to do a lot more to get it working, it was mostly a case of a really old docker installation locally. Tests are coming up green both locally and under github actions. Some things (issues with `xerces` & `gnupg`) were fixed by @aelkiss in https://github.com/hathitrust/feed/commit/8b4115dcd92cde5fefb28dd7d1eb99f7fe976c19.

Also had to mess around with my ssh keys so I could start using the ssh protocol, because you are no longer allowed to push to workflows with https for some (security?) reason. That's not in the PR either.

## Actual changes

Changing `docker-compose` to hyphenless `docker compose` and removing a few warnings from the tests.

## To review

(You may have to stop or prune existing services.)

```
cd /tmp
git clone -b validate-cache-usr-local git@github.com:hathitrust/feed ./feed_clean
cd /tmp/feed_clean
docker build
docker compose run test
```

## Expected results: 

It'll take several minutes to build and test.

Expect to see some warnings, but ultimately a message that tests passed:
`Result: PASS`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,9 +17,9 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Build docker image
-        run: docker-compose --file docker-compose.test.yml build
+        run: docker compose --file docker-compose.test.yml build
 
       - name: Run tests
-        run: docker-compose --file docker-compose.test.yml run sut
+        run: docker compose --file docker-compose.test.yml run sut
         env:
           GITHUB_TOKEN: ${{ secrets.github_token }}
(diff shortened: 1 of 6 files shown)


aelkiss (approved):
Nice. I like the `no warnings 'redefine'` especially - that's always an annoying thing to see, but not one I've (obviously) ever bothered to take care of

mwarin:
deleting an old merged branch

PR #111, opened 2024-01-30
DEV-989 feed: processing jpgs & pngs in digifeed packages

mwarin:
This goes hand in hand with: https://github.com/hathitrust/feed_internal/pull/34

### DIFF
--- a/lib/HTFeed/ModuleValidator.pm
+++ b/lib/HTFeed/ModuleValidator.pm
@@ -2,15 +2,15 @@ package HTFeed::ModuleValidator;
 
 use warnings;
 use strict;
+
 use Carp;
-use XML::LibXML;
-use Log::Log4perl qw(get_logger);
 use Data::Dumper qw(Dumper);
+use HTFeed::Config qw(get_config);
+use HTFeed::XMLNamespaces qw(register_namespaces);
+use Log::Log4perl qw(get_logger);
+use XML::LibXML;
 
 use base qw(HTFeed::XPathValidator);
-
-use HTFeed::XMLNamespaces qw(register_namespaces);
-use HTFeed::Config qw(get_config);
 
 =head1 NAME
 
@@ -30,10 +30,9 @@ HTFeed::ModuleValidator
 
 	my $context_node = $xpc->findnodes("$repInfo/$format"."Metadata");
 	my $validator = HTFeed::ModuleValidator::JPEG2000_hul->new(xpc => $xpc, qlib => $querylib);
-	if ($validator->validate){
+	if ($validator->validate) {
 		# SUCCESS code...
-	}	
-	else{
+	} else {
 		my $errors = $validator->getErrors;
 		# FAILURE code...
 	}
@@ -75,22 +74,20 @@ sub new {
     $object->{filename} =~ /\.([0-9a-zA-Z]+)$/;
     my $file_ext = $1;
 
-    # get module validator list from volume, set class accordingly
-    $class =
-      $object->{volume}->get_nspkg()->get('module_validators')->{$file_ext}
-      or croak "Don't know how to validate file extension $file_ext";
+    my $module_validators = $object->{volume}->get_nspkg()->get('module_validators');
+    defined $module_validators or croak("No module_validators found for " . $object->{filename});
+    my $ext_validator = $module_validators->{$file_ext};
+    defined $ext_validator or croak("None of the module_validators match file_ext $file_ext");
 
-    bless( $object, $class );
-
+    bless( $object, $ext_validator );
     $object->_xpathInit();
     $object->_set_validators();
 
-    my $overrides =
-      $object->{volume}->get_validation_overrides($class);
+    my $overrides = $object->{volume}->get_validation_overrides($ext_validator);
     while ( my ( $k, $v ) = each(%$overrides) ) {
         $object->{validators}{$k}{valid} = $v;
         $object->{validators}{$k}{desc} = $k if not defined $object->{validators}{$k}{desc};
-        $object->{validators}{$k}{detail} = "Package type specific - see $class" if not defined $object->{validators}{$k}{detail};
+        $object->{validators}{$k}{detail} = "Package type specific - see $ext_validator" if not defined $object->{validators}{$k}{detail};
     }
 
     return $object;
--- a/lib/HTFeed/Stage/ImageRemediate.pm
+++ b/lib/HTFeed/Stage/ImageRemediate.pm
@@ -2,18 +2,22 @@ package HTFeed::Stage::ImageRemediate;
 
 use strict;
 use warnings;
+
 use base qw(HTFeed::Stage::JHOVE_Runner);
-use List::Util qw(max min);
-use POSIX qw(ceil);
-use HTFeed::Config qw(get_config);
-use Log::Log4perl qw(get_logger);
-use File::Copy;
+
 use Carp;
-use HTFeed::XMLNamespaces qw(register_namespaces);
-use Encode qw(decode);
-use File::Basename qw(basename);
 use Date::Manip;
+use Encode qw(decode);
+use File::Basename qw(basename fileparse);
+use File::Copy;
+use HTFeed::Config qw(get_config);
+use HTFeed::XMLNamespaces qw(register_namespaces);
 use Image::ExifTool;
+use List::Util qw(max min);
+use Log::Log4perl qw(get_logger);
+use POSIX qw(ceil);
+
+use Data::Dumper qw(Dumper);
 
 =head1 NAME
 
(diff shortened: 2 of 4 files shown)


lib/HTFeed/ModuleValidator.pm
```diff
@@ -75,22 +74,20 @@ sub new {
     $object->{filename} =~ /\.([0-9a-zA-Z]+)$/;
     my $file_ext = $1;
 
-    # get module validator list from volume, set class accordingly
```
mwarin, line 78: Should be equivalent but hopefully easier to read.

lib/HTFeed/ModuleValidator.pm
```diff
@@ -2,16 +2,16 @@ package HTFeed::ModuleValidator;
 
 use warnings;
 use strict;
+
 use Carp;
-use XML::LibXML;
```
mwarin, line 6: sorted to make more readable

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -2,18 +2,22 @@ package HTFeed::Stage::ImageRemediate;
 
 use strict;
```
mwarin, line 3: sorted for readability

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -369,7 +376,152 @@ sub _remediate_tiff {
       && $self->repair_tiff_exiftool( $infile, $outfile, $self->{newFields} );
 
     return $ret;
+}
+
+# Extract metadata from jpg/png and write to the corresponding tiff.
```
mwarin: New method. Delegates to the relevant `extract_$ext_metadata` writes extracted fields to output file.

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -369,7 +376,152 @@ sub _remediate_tiff {
...
+    }
+    my $exif_write_status = $exifTool->WriteInfo($outfile);
+    unless ($exif_write_status == 1) {
+	get_logger()->trace("Failed EXIF write to $outfile");
+    }
+}
+
+# Build and return a hash of jpg metadata.
+sub extract_jpg_metadata {
+    my $olf = shift; # ref to $self->{oldFields}, a hash of exiftool data
```
mwarin, line 419: Here I mostly just don't know enough about image metadata to figure it out.

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -891,33 +1070,41 @@ sub convert_tiff_to_jpeg2000 {
...
             or $self->{oldFields}->{'IFD0:BitsPerSample'} eq '8 8 8')) {
         $imagemagick_cmd .= qq( -type TrueColor)
-    } elsif($self->{oldFields}->{'IFD0:BitsPerSample'} eq '8' 
+    } elsif($self->{oldFields}->{'IFD0:BitsPerSample'} eq '8'
             and $self->{oldFields}->{'IFD0:SamplesPerPixel'} eq '1') {
         $imagemagick_cmd .= qq( -type Grayscale -depth 8)
     }
 
-    # strip 
+    my $magick_compress_cmd = "$imagemagick_cmd -compress None $infile -strip $infile.unc.tif";
```
mwarin, line 998: rewrote for readability

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -891,33 +1070,41 @@ sub convert_tiff_to_jpeg2000 {
...
 
-    system(qq($grk_compress -i "$infile.unc.tif" -o "$outfile" -p RLCP -n $levels -SOP -EPH -M 62 -I > /dev/null 2>&1))
-
-      and $self->set_error(
-        "OperationFailed",
-        operation => "grk_compress",
-        file      => $infile,
-        detail    => "grk_compress returned $?"
-      );
+    my $grk_compress_cmd = qq($grk_compress -i "$infile.unc.tif" -o "$outfile" -p RLCP -n $levels -SOP -EPH -M 62 -I > /dev/null 2>&1);
```
mwarin, line 1016: rewrote for readability

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -186,86 +203,82 @@ sub stage_info {
...
 }
 
+
 sub _remediate_tiff {
     my $self                     = shift;
     my $infile                   = shift;
     my $outfile                  = shift;
     my $force_headers            = ( shift or {} );
     my $set_if_undefined_headers = shift;
+    my $check_jhove_status       = shift;
```
aelkiss: I think this might have been a direction that we didn't end up using? Not 100% sure

mwarin: Yes, this was an attempt at a solution we ended up without. Fully gone now.

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -369,7 +370,107 @@ sub _remediate_tiff {
...
+    };
+
+    return $h;
+}
+
+# Build & return a hash of metadata extracted from a png.
+sub extract_png_metadata {
+    my $olf = shift; # ref to $self->{oldFields}, a hash of exiftool data
+    my $h = {
+	'XMP-tiff:SamplesPerPixel' => $olf->{'PNG:BitDepth'},
```
aelkiss: We shouldn't need to copy ImageHeight, ImageWidth, or SamplesPerPixel

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -369,7 +370,107 @@ sub _remediate_tiff {
...
+	'XMP-tiff:ResolutionUnit'  => "cm"
+    };
+
+    # $exifTool->SetNewValue does NOT like PNG's compression value.
+    # $h->{'XMP-tiff:Compression'} = $olf->{'PNG:Compression'};
+    # Results in:
+    ## Error setting new tag XMP-tiff:Compression
+    ## Deflate/Inflate: Can't convert XMP-tiff:Compression (not in PrintConv)
+
+    # Several metadata fields can/must be inferred from PNG:ColorType:
```
aelkiss: We should not need to copy any of these

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -369,7 +370,107 @@ sub _remediate_tiff {
...
+    unless ($exif_write_status == 1) {
+	get_logger()->trace("Failed EXIF write to $outfile");
+    }
+}
+
+# Build and return a hash of jpg metadata.
+sub extract_jpg_metadata {
+    my $olf = shift; # ref to $self->{oldFields}, a hash of exiftool data
+    my $h = {
+	'IFD0:ModifyDate'         => $olf->{'System:FileModifyDate'},
```
aelkiss: As below I don't think we need to copy `ImageHeight`, `ImageWidth`, `BitsPerSample`. I think `dc:source`, `IFD0:ModifyDate`, `tiff:DateTime` probably aren't necessary because we normalize those later and use whatever information we can find (capture date from the METS or fallback to filesystem timestamp if need be, I think).

lib/HTFeed/Stage/ImageRemediate.pm
aelkiss: We should try to get resolution from `PNG:PixelsPerUnitX`, `PNG:PixelsPerUnitY`, `PNG:PixelUnits`

aelkiss (changes requested):
Reformatting looks fine. There are certainly more opportunities for refactoring, unit testing within `ImageRemediate`, and rethinking the relationship and flow of code between `ImageRemediate` subclasses and this class, but I don't think we need to worry about that now.

The change I do think we need is some adjustment to headers we're extracting -- in particular we shouldn't copy the image height, width, color depth, etc, as this will all get carried forward when ImageMagick converts to TIFF -- but we do need to make sure to try to get the resolution info from PNGs.

lib/HTFeed/Stage/ImageRemediate.pm
mwarin: I made the extract_x_metadata methods a little tighter, so now they only get the resolution-related metadata:
```
'IFD0:ResolutionUnit'    
'IFD0:XResolution'       
'IFD0:YResolution'       
'XMP-tiff:ResolutionUnit'
'XMP-tiff:XResolution'   
'XMP-tiff:YResolution'   
```

lib/HTFeed/Stage/ImageRemediate.pm
mwarin: Done. 
PNG puts them in the `'PNG-pHYs:...'` block, and seems to default to meters as the unit.

lib/HTFeed/Stage/ImageRemediate.pm
mwarin: Done, they are gone.

lib/HTFeed/Stage/ImageRemediate.pm
mwarin: OK, not getting any of them.

aelkiss (approved):
This is looking good to me at this point (assuming that tests in feed_internal pass when running against this branch) Please feel free to do a merge & release when ready.

mwarin:
deleting an old merged branch

PR #113, opened 2024-03-20
DEV-1075: add HTFeed::JobMetrics

mwarin:
Originating ticket: https://hathitrust.atlassian.net/browse/DEV-1075

## What?

This PR adds the class `HTFeed::JobMetrics` that can be used to gather metrics on the various stages of a job, such as number of items downloaded, number of bytes downloaded and amount of time spent downloading (as well as the other job stages, not just download).

`HTFeed::JobMetrics` specifies which kinds of metrics can be used, knows how to update them, pretty print and clear metrics. It has its own minimal CLI (see `perl lib/HTFeed/JobMetrics.pm --usage`)

## Why?

We want a better understanding of how much time is spent on different parts of the Ingest workflow.

## How?

`--help` or `--usage` should provide a short usage page.

Each measured job reports how long it took to `HTFeed::JobMetrics`, which passes it on to `Prometheus::Tiny::Shared` which saves the data point in a `Hash::SharedMem`.

The `Hash::SharedMem` needs to live somewhere on the filesystem. For now, it defaults to a subdir of `/tmp/`:
```perl
$self->{file} = "/tmp/HTFeed::JobMetrics::Data";
if (defined $ENV{'HTFEED_JOBMETRICS_DATA_DIR'}) {
    $self->{file} = $ENV{'HTFEED_JOBMETRICS_DATA_DIR'};
}
$self->{prom} = Prometheus::Tiny::Shared->new(filename => $self->{file});
```
... but that can be changed by setting `$ENV{'HTFEED_JOBMETRICS_DATA_DIR'}` to some other path.

Metrics will generally change when jobs are run. 
For ease of testing, they can be viewed/manipulated directly via commandline:

```
$ docker compose run --rm test bash

perl lib/HTFeed/JobMetrics.pm --clear
  Clearing data...
  Data cleared.
  No operation defined, exiting.

perl lib/HTFeed/JobMetrics.pm --operation match --value collate
  running operation match ...
  # TYPE ingest_collate_bytes counter
  # TYPE ingest_collate_items counter
  # TYPE ingest_collate_seconds counter

perl lib/HTFeed/JobMetrics.pm --operation inc --metric ingest_collate_items
  running operation inc ...

perl lib/HTFeed/JobMetrics.pm --operation add --metric ingest_collate_bytes --value 2460
  running operation add ...

perl lib/HTFeed/JobMetrics.pm --operation match --value collate
  running operation match ...
  # TYPE ingest_collate_bytes counter
  ingest_collate_bytes 2460
  # TYPE ingest_collate_items counter
  ingest_collate_items 1
  # TYPE ingest_collate_seconds counter
```

Get all metrics with:
```
$ docker compose run --rm test perl lib/HTFeed/JobMetrics.pm --pretty
```

## Limitations

Note that if you call from outside docker, and have not set `$ENV{'HTFEED_JOBMETRICS_DATA_DIR'}`, then the metrics data will default to being stored in the container's `/tmp`, and that data will not persist between commands.

Far from every operation that could be measured is being measured, but the size of this PR is getting out of hand and those missing pieces could be added incrementally after we've proven whether this concept works in production.

## Tests

`docker compose run --rm test perl t/job_metrics.t` to run just the new tests.
`docker compose run --rm test` to run all tests.

## Where?

These are some of the main changes per file:

* `lib/HTFeed/JobMetrics.pm` new class, the main innovation in this PR.
* `t/job_metrics.t` tests for the new class.
* `lib/HTFeed/QueueRunner.pm` uses JobMetrics to observe `_items` and `_seconds` metrics for each job stage in `run_job_sequence` (this is a big and important part of this PR so this needs to work)
* `Makefile.pl` getting required perl module `Prometheus::Tiny::Shared`.
* `lib/HTFeed/Stage.pm` giving a `JobMetrics` object (`$self->{job_metrics}`) to all `Stage` objects and their heirs.

The rest of the `Files changed` are about using `HTFeed::JobMetrics`, plus some cleanup/cosmetics.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,8 +60,6 @@ RUN echo "deb [signed-by=/usr/share/keyrings/hathitrust-archive-keyring.gpg] htt
 
 RUN apt-get update && apt-get install -y grokj2k-tools
 
-RUN cpan -f -i Net::AMQP::RabbitMQ
-
 COPY etc/imagemagick-policy.xml /etc/ImageMagick-6/policy.xml
 
 COPY etc/jhove-auto-install.xml /tmp/jhove-auto-install.xml
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -24,6 +24,7 @@ WriteMakefile(
     'Mouse'               => 0,
     'Net::AMQP::RabbitMQ' => 0,
     'Net::Prometheus'     => 0,
+    'Prometheus::Tiny::Shared' => 0,
     'ProgressTracker'     => 0,
     'Readonly'            => 0,
     'Roman'               => 0,
--- a/bin/validate_volume.pl
+++ b/bin/validate_volume.pl
@@ -55,35 +55,33 @@ if ($one_line){
 }
 
 unless ($realmeta) {
+    no warnings 'redefine';
 
-  *HTFeed::Volume::get_sources = sub {
-    return ( 'ht_test','ht_test','ht_test' );
-  };
+    *HTFeed::Volume::get_sources = sub {
+	return ('ht_test', 'ht_test', 'ht_test');
+    };
 
-  # use faked-up marc in case it's missing
-
-  *HTFeed::SourceMETS::_get_marc_from_zephir = sub {
-    my $self = shift;
-    my $marc_path = shift;
-
-    my $identifier = $self->{volume}->get_identifier();
-
-    if (not HTFeed::Stage::Download::download($self,
-      url => "http://zephir.cdlib.org/api/item/" . $self->{volume}->get_identifier(),
-      path => dirname($marc_path),
-      filename => basename($marc_path),
-      not_found_ok => 1)) {
-
-
-      HTFeed::Stage::Download::download($self,
-        url => "http://zephir.cdlib.org/api/item/mdp.39015039746220",
-        path => dirname($marc_path),
-        filename => basename($marc_path),
-        not_found_ok => 1);
-      
-    }
-
-  };
+    # use faked-up marc in case it's missing
+    *HTFeed::SourceMETS::_get_marc_from_zephir = sub {
+	my $self = shift;
+	my $marc_path = shift;
+	my $downloaded = HTFeed::Stage::Download::download(
+	    $self,
+	    url => "http://zephir.cdlib.org/api/item/" . $self->{volume}->get_identifier(),
+	    path => dirname($marc_path),
+	    filename => basename($marc_path),
+	    not_found_ok => 1
+	);
+	if (not $downloaded) {
+	    HTFeed::Stage::Download::download(
+		$self,
+		url => "http://zephir.cdlib.org/api/item/mdp.39015039746220",
+		path => dirname($marc_path),
+		filename => basename($marc_path),
+		not_found_ok => 1
+	    );
+	}
+    };
 }
 
 pod2usage(-msg => 'must specify package type with -p or -d') if not defined $default_packagetype and not defined $dot_packagetype;
--- a/lib/HTFeed/Job.pm
+++ b/lib/HTFeed/Job.pm
@@ -1,12 +1,10 @@
 package HTFeed::Job;
 
-#use Moose;
-use Mouse;
-use HTFeed::Volume;
-use HTFeed::Config;
 use Carp;
-
+use HTFeed::Config;
+use HTFeed::Volume;
 use Log::Log4perl qw(get_logger);
+use Mouse;
 
 has [qw(pkg_type namespace id)]         => (is => 'ro', isa => 'Str',     lazy_build => 1);
 has 'status'                            => (is => 'ro', isa => 'Str',     required => 1, default => 'ready');
@@ -17,6 +15,7 @@ has 'new_status'                        => (is => 'ro', isa => 'Str',     init_a
 has '_release'                          => (is => 'ro', isa => 'Bool',    init_arg => undef, lazy_build => 1);
 has 'volume'                            => (is => 'ro', isa => 'Object',                     lazy_build => 1);
 has 'stage'                             => (is => 'ro', isa => 'Object',  init_arg => undef, lazy_build => 1);
+
 
 =head1 NAME
 
@@ -54,22 +53,40 @@ HTFeed::Job->new(   volume => $volume,
 sub update{
     my $self = shift;
 
-    my $stage = $self->stage;
-    my $fail = $stage->failed;
-    my $new_status = $self->new_status;
+    my $stage       = $self->stage;
+    my $fail        = $stage->failed;
+    my $new_status  = $self->new_status;
 
     ## TODO: make this a class global or see if it can be better accessed with YAML::Config, etc.
     ## i.e. put it somwhere else, but preferably somthing tidy
     my %release_states = map {$_ => 1} @{get_config('release_states')};
-
     my $release = $self->_release;
 
-    get_logger()->info( 'StageSucceeded', objid => $self->id, namespace => $self->namespace, stage => $self->stage_class, detail => $stage->success_info() )
-        if (!$fail);
-    get_logger()->info( 'StageFailed', objid => $self->id, namespace => $self->namespace, stage => $self->stage_class, detail => 'fatal=' . ($new_status eq 'punted' ? '1' : '0') )
-        if ($fail);
+    if ($fail) {
+	get_logger()->info(
+	    'StageFailed',
+	    objid     => $self->id,
+	    namespace => $self->namespace,
+	    stage     => $self->stage_class,
+	    detail    => 'fatal=' . ($new_status eq 'punted' ? '1' : '0')
+	);
+    } else {
+	get_logger()->info(
+	    'StageSucceeded',
+	    objid     => $self->id,
+	    namespace => $self->namespace,
+	    stage     => $self->stage_class,
+	    detail    => $stage->success_info()
+	);
+    }
 
-    &{$self->{callback}}($self->namespace, $self->id, $new_status, $release, $fail);
+    &{$self->{callback}}(
+	$self->namespace,
+	$self->id,
+	$new_status,
+	$release,
+	$fail
+    );
 
     return;
 }
@@ -101,9 +118,8 @@ sub clean{
     my $stage = $self->stage;
 
     $stage->clean();
-    
     $stage->clean_punt() if ($self->new_status eq 'punted');
-    
+
     return;
 }
 
@@ -111,17 +127,19 @@ sub clean{
 around BUILDARGS => sub {
     my $orig  = shift;
     my $class = shift;
-    
+
     # exactly 6 args to construct w/o hash style args
     # 10-12 with hash style args
     if ( @_ == 6 ) {
         my ($pkg_type, $namespace, $id, $status, $failure_count, $callback) = @_;
-        return $class->$orig(pkg_type => $pkg_type,
-                             namespace => $namespace,
-                             id => $id,
-                             status => $status,
-                             failure_count => $failure_count,
-                             callback => $callback);
+        return $class->$orig(
+	    pkg_type      => $pkg_type,
+	    namespace     => $namespace,
+	    id            => $id,
+	    status        => $status,
+	    failure_count => $failure_count,
+	    callback      => $callback
+	);
     }
     else {
         return $class->$orig(@_);
@@ -131,12 +149,31 @@ around BUILDARGS => sub {
 sub BUILD{
     my $self = shift;
     my $args = shift;
-    
-    return
-        if( (!$self->has_pkg_type && !$self->has_namespace && !$self->has_id and $self->has_volume) or
-            ($self->has_pkg_type && $self->has_namespace && $self->has_id and !$self->has_volume) );
 
-    croak 'Error instantiating Job: Must specify namespace, id, and packagetype, OR specify volume object. Must not specify both.';
+    my $cond1 = (
+	!$self->has_pkg_type  &&
+	!$self->has_namespace &&
+	!$self->has_id        and
+	$self->has_volume
+    );
+
+    my $cond2 = (
+	$self->has_pkg_type  &&
+	$self->has_namespace &&
+	$self->has_id        and
+	!$self->has_volume
+    );
+
+    # Desired outcome.
+    return if ($cond1 or $cond2);
+
+    # Undesired outcome.
+    croak join(
+	" ",
+	"Error instantiating Job:",
+	"Must specify namespace, id, and packagetype, OR specify volume object.",
+	"Must not specify both."
+    );
 }
 
 sub _build_volume{
@@ -150,15 +187,15 @@ sub _build_volume{
 }
 
 sub _build_namespace {
-	my $self = shift;
+    my $self = shift;
     return $self->volume->get_namespace;
 }
 sub _build_pkg_type {
-	my $self = shift;
+    my $self = shift;
     return $self->volume->get_packagetype;
 }
 sub _build_id {
-	my $self = shift;
+    my $self = shift;
     return $self->volume->get_objid;
 }
 
@@ -172,23 +209,24 @@ sub _build_stage_class{
 
 sub _build_stage{
     my $self = shift;
-    
+
     my $class = $self->stage_class;
     my $volume = $self->volume;
-    
+
     return eval "$class->new(volume => \$volume)";
 }
 
 sub _build_new_status{
     my $self = shift;
-    
+
     my $stage = $self->stage;
-    
     my $success = $stage->succeeded;
-    my $new_status = $success ? 
+    my $new_status = $success ?
         $stage->get_stage_info('success_state') : $stage->get_stage_info('failure_state');
-    $new_status = 'punted'
-        if((! $success) and ($self->failure_count >= get_config('failure_limit')));
+
+    if ( (! $success) and ($self->failure_count >= get_config('failure_limit')) ) {
+	$new_status = 'punted';
+    }
 
     # punt if next status is undefined
     $new_status = 'punted' unless $new_status;
@@ -225,19 +263,33 @@ sub run_job {
     my $force_failed_status = shift;
 
     my $stage;
-
+    
     eval {
         $stage = $job->stage;
-        
-        get_logger()->info( 'RunStage', objid => $job->id, namespace => $job->namespace, stage => $job->stage_class );
+
+        get_logger()->info(
+	    'RunStage',
+	    objid     => $job->id,
+	    namespace => $job->namespace,
+	    stage     => $job->stage_class
+	);
 
         $job->volume->reset();
+
+	
         $stage->run();
+
     };
 
     my $err = $@;
     if ( $err and $err !~ /STAGE_ERROR/ and $err !~ /VOLUME_ERROR/) {
-        get_logger()->error( 'UnexpectedError', objid => $job->id, namespace => $job->namespace, stage => $job->stage_class, detail => $err );
+        get_logger()->error(
+	    'UnexpectedError',
+	    objid     => $job->id,
+	    namespace => $job->namespace,
+	    stage     => $job->stage_class,
+	    detail    => $err
+	);
     }
 
     # handle fake status set in unit tests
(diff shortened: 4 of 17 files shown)


lib/HTFeed/JobMetrics.pm
```diff
@@ -0,0 +1,310 @@
+package HTFeed::JobMetrics;
```
mwarin, line 1: @aelkiss if you want to start looking before I'm 100% done with the PR you could start with this class.

lib/HTFeed/JobMetrics.pm
```diff
@@ -0,0 +1,330 @@
...
+
+# irrevocably delete job metrics
+sub clear {
+    my $self = shift;
+    $self->{prom}->clear;
+}
+
+# We'll mostly get sizes in b (from perl's -s operator),
+# and want to show them in kb.
+sub b_to_kb {
```
aelkiss: Prometheus suggests recording metrics as bytes rather than kilobytes; may be worth reviewing:

https://prometheus.io/docs/practices/naming/

mwarin: Alright, that simplifies some things.

mwarin: How do you feel about seconds/milliseconds though?

... that page:
> should use base units (e.g. seconds, bytes, meters - not milliseconds, megabytes, kilometers)

aelkiss: I would use seconds and report a decimal/floating point value

lib/HTFeed/JobMetrics.pm
```diff
@@ -0,0 +1,320 @@
...
+# $jm->add("Collate_bytes", 2460) ;     # for counters    (+= x)
+# $jm->observe("Download_ms", 220); # for historgrams (+= x)
+# print $jm->pretty;
+### list available metrics
+# $jm->list_metrics
+^;
+
+# If called from commandline:
+if (!caller) {
+    my $jm = HTFeed::JobMetrics->new();
```
aelkiss: The idea we might run this from the command line in a separate process makes me wonder if we could use this across different imgsrv processes not running under plack/psgi (e.g. including the ones that run under CGI...) :thinking: worth experimenting with there

aelkiss: I also wonder if we might be able to pull this out to use for imgsrv as well or otherwise deduplicate common functionality. I will aim to share draft PRs for imgsrv/mdp-lib tomorrow

lib/HTFeed/JobMetrics.pm
```diff
@@ -0,0 +1,320 @@
...
+
+    if (!@ARGV || grep { /--(help|usage)/ } @ARGV) {
+	print $USAGE;
+	exit;
+    }
+
+    while (my $arg = shift @ARGV) {
+	if ($arg =~ /^--pretty$/) {
```
aelkiss: if we retain this functionality in the final version I would want to see it use `Getopt::Long`

lib/HTFeed/JobMetrics.pm
```diff
@@ -0,0 +1,320 @@
...
+}
+
+sub _setup_metrics {
+    my $self = shift;
+    my $prom = $self->{prom};
+
+    # Each element x in this array gets turned into 3 prom declarations:
+    # x_ms, x_bytes and x_items
+    my @measurable_stages = (
+	'Collate',
```
aelkiss: To better fit the naming scheme for prometheus metrics I might suggest lowercasing these and prepending `ingest_` or `ingest_stage_`

mwarin: New naming scheme uses `ingest_` prefix and lowercases the stage name. What used to be `Pack_ms` is now, after all the changes, `ingest_pack_seconds`.

lib/HTFeed/JobMetrics.pm
mwarin: OK, now measuring bytes and seconds.

lib/HTFeed/JobMetrics.pm
mwarin: I'll look at that. 

It seemed excessive to use Getopt at first, then the CLI code grew into something that definitely needs to be managed... so yeah, agree.

mwarin: Now using `GetOpt::Long`.

lib/HTFeed/Job.pm
```diff
@@ -131,12 +149,31 @@ around BUILDARGS => sub {
 sub BUILD{
     my $self = shift;
     my $args = shift;
-    
-    return
-        if( (!$self->has_pkg_type && !$self->has_namespace && !$self->has_id and $self->has_volume) or
-            ($self->has_pkg_type && $self->has_namespace && $self->has_id and !$self->has_volume) );
 
-    croak 'Error instantiating Job: Must specify namespace, id, and packagetype, OR specify volume object. Must not specify both.';
+    my $cond1 = (
```
aelkiss, line 153: I appreciate all the tidyings; I think it does help clarify what's going on and make it more readable. I think you were going to start looking at [_Tidy First_](https://learning.oreilly.com/library/view/tidy-first/9781098151232/). That might help in terms of thinking about structuring commits & pull requests to separate tidying/refactoring from behavior change. (No need to change anything in this PR, to be clear!)

lib/HTFeed/JobMetrics.pm
```diff
@@ -0,0 +1,352 @@
...
+	exit(0);
+    }
+}
+# end commandline stuff
+
+sub new {
+    my $class = shift;
+    my $self  = {};
+
+    # TODO: Need to determine best place to store metrics
```
aelkiss: I would say keep it in `/tmp` (NFS via `/htprep/` would likely cause a variety of problems for this) but would suggest using a name like `/tmp/htfeed-jobmetrics-data` (i.e. without colons). It's probably fine as-is but punctuation in filenames always makes me antsy

lib/HTFeed/JobMetrics.pm
```diff
@@ -0,0 +1,352 @@
...
+
+=item --pretty
+
+Full, somewhat readable, accounting of known metrics and their values.
+
+=cut
+
+sub pretty {
+    my $self = shift;
+    $self->{prom}->format;
```
aelkiss, line 173: I think the idea in this case is that the pod that's being scraped could upon request access the same shared file and then call this method when it wants to render the metrics?

lib/HTFeed/JobMetrics.pm
```diff
@@ -0,0 +1,352 @@
...
+}
+
+# "private" from here on, indicated with _leading_underscores
+# in method names
+
+# Check that only valid metrics are used,
+# or make an intelligent complaint.
+# e.g. $jm->_valid_metric("Unpack_seconds") # -> 1
+# e.g. $jm->_valid_metric("invalid metric name") # -> warn & 0
+sub _valid_metric {
```
aelkiss, line 299: I'm waffling on whether it's necessary to do error checking here. If we did report an "invalid" metric then I believe `Prometheus::Tiny::Shared` will happily report it, and we would scrape it and have it in our prometheus instance. We might then have trouble discovering it was there if it wasn't named as we expected; it might be easier to find the misnamed metric in Prometheus than it would be to find from the logs that something was wrong. If we didn't have this error checking here, I wouldn't say that we needed it; since it is, I think it's OK to keep, and we can see if we find it helpful or a hindrance later on.

lib/HTFeed/JobMetrics.pm
```diff
@@ -0,0 +1,352 @@
...
+    # Prefix all metric names with:
+    my $pfx = "ingest";
+
+    # Each element x in this array gets turned into 3 prom declarations:
+    # x_seconds, x_bytes and x_items
+    # Names are based on the downcased class name of the stage.
+    my @measurable_stages = (
+	'collate',
+	'download',
+	'download_dropbox',
```
aelkiss, line 327: A couple thoughts here:
* it might be helpful to have 'dropbox','google','ia' as labels rather than as separate metrics (then we could sum across all of them, and maybe not need to separately record download_bytes)?
* for 'seconds' it might be helpful to report a single `ingest_run_time_seconds` metric with a label based on the stage - again then we can sum across all of them. It might also be helpful to then use these as labels for a `ingest_processed_bytes` metric with a label based on the stage, and the same for `ingest_processed_items`. We could then perhaps look at something like an aggregation across average bytes per item by stage, or average seconds per item by stage.

lib/HTFeed/PackageType/Simple/Download.pm
```diff
@@ -3,56 +3,73 @@ package HTFeed::PackageType::Simple::Download;
...
+	    detail    => "$sip_directory could not be created"
+	);
+    }
+
+    eval {
+	my $rclone = HTFeed::Rclone->new;
+	$rclone->copy($url, $sip_directory);
+	my $download_size;
+	find (sub { -f and ( $download_size += -s ) }, $sip_directory);
+	$self->{jm}->add("Download_bytes", $download_size);
```
aelkiss: Is there a difference between `jm` and `job_metrics`? If not, I'd prefer to consistently use `job_metrics`. If so, I think `jm` could still use a more intention-revealing name.

lib/HTFeed/QueueRunner.pm
```diff
@@ -104,19 +106,30 @@ sub validate_job {
...
+    while($job){
+	get_logger()->info(
+	    "next job: " . $job->namespace . "." . $job->id . " " . $job->stage_class
+	);
+	my $jobtype = (split("::", $job->stage_class))[-1];
+	my $t_start = Time::HiRes::time();
+	$job->run_job($self->{clean});
+	my $t_end = Time::HiRes::time();                  # gives milliseconds,
+	my $t_delta_seconds = ($t_end - $t_start) * 1000; # multiply to seconds
+	$self->{job_metrics}->add($jobtype."_seconds", $t_end - $t_start);
```
aelkiss: Would it make sense to push this logic into `Job` rather than having it here? If we do keep it here I would suggest extracting it as a separate method though.

lib/HTFeed/QueueRunner.pm
```diff
@@ -104,19 +106,30 @@ sub validate_job {
...
+	$self->{job_metrics}->inc($jobtype."_items");
+	if ($jobtype eq "Download") {
+	    # TODO: for job metrics, not sure what is best way of determining
+	    # if a download job is for ia/google/dropbox
+	}
+	# TODO: for job_metrics, not sure what the best way is to get filesize from job
```
aelkiss, line 134: I think we could incorporate this into the various download stages and record it there where we have the info.

lib/HTFeed/Stage.pm
```diff
@@ -23,11 +23,14 @@ sub new {
...
     }
 
+    # No class inheriting from HTFeed::Stage needs to instantiate
+    # its own $self->{jm} for storing JobMetrics.
     my $self = {
         volume => undef,
         @_,
         has_run => 0,
         failed  => 0,
+	jm      => HTFeed::JobMetrics->new
```
aelkiss: Do the tests cover the case of multiple instances of `HTFeed::JobMetrics` writing to the same file in the same process? (e.g. we have one in `QueueRunner`, then each individual `Stage` gets one..)

If this works, it might point to an approach that could work in imgsrv/mdp-lib apps as well in terms of sidestepping the need to pass around a single object or to keep track of one globally.

If this doesn't work as-is, we could perhaps change `HTFeed::JobMetrics` to implement a kind of singleton pattern where you do like `HTFeed::JobMetrics->instance` that initializes the single new one if we need it.

lib/HTFeed/Stage/Download.pm
```diff
@@ -64,8 +63,9 @@ sub download{
 
     if( $response->is_success() ){
         my $size = (-s $pathname);
+	$self->{jm}->add("Download_bytes", $size);
```
aelkiss: Could add a label here based on the hostname we are downloading from?

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -116,6 +116,8 @@ sub remediate_image {
 	return $self->_remediate_tiff( $oldfile, @_ );
     }
 
+    $self->{jm}->inc("ImageRemediate_images");
```
aelkiss: Could potentially add a label here (and elsewhere in imageremediate) based on the kind of image we are remediating (tiff vs. jpeg2000 vs. png vs. jpeg?)

lib/HTFeed/Stage/Pack.pm
```diff
@@ -73,24 +76,43 @@ sub run{
...
+	    );
             return;
-        }        
+        }
     }
 
+    # Get size after zipping and report size delta as a job metric for Pack.
+    my $end_size = -s $zip_path;
+    my $delta_size = $start_size - $end_size;
+    $self->{jm}->add("Pack_bytes", $delta_size);
```
aelkiss: Might rather record the number of (uncompressed) bytes read and the (compressed) bytes written? The delta size is likely to generally be rather small and I'm not sure terribly meaningful

aelkiss (commented):
Lots of good stuff here as a starting point.

One thing to address right now is a couple additional tests:
* something involving at least one `Stage` that verifies that it records the expected metrics (not necessarily the exact values, just some nonzero value)
* an integration test involving `QueueRunner` and multiple `Stage`s and again verifies we see non-zero values for expected metrics.

as well as the issue of `jm` vs `job_metrics`.

A couple of things to think about later on:
* consider whether metrics tracking belongs in `QueueRunner` vs. `Job`
* Would definitely recommend thinking more about specific metric names and labels, although that also needs to be informed by specific use cases; we can talk about that more
* Would definitely want metrics in `HTFeed::Stage::Collate` and probably its more specific operations (https://github.com/hathitrust/feed/blob/main/lib/HTFeed/Stage/Collate.pm#L78-L86)
* We may also want to add some metrics in `feed_internal` - one specific area might be this operation https://github.com/hathitrust/feed_internal/blob/main/lib/HTFeed/PackageType/Google/Unpack.pm#L42

I would suggest making follow-up subtasks for these areas, though -- I don't think they need to block merging.

lib/HTFeed/JobMetrics.pm
mwarin: Yep. Changed default path to `/tmp/htfeed-jobmetrics-data`, removed the `#todo`.

lib/HTFeed/JobMetrics.pm
mwarin: Ah. Yes. I added a sentence to the sub's pod to reflect that.
Maybe it would make sense to name it "export" or "scrape" but what it really is is a pretty-printer.

lib/HTFeed/JobMetrics.pm
mwarin: Honestly, this was mostly a crutch I used during development. I agree that it isn't strictly necessary, more for catching typos than anything else. I say keep until it becomes annoying. (When we start adding metrics we'll find out.)

lib/HTFeed/QueueRunner.pm
mwarin: Maybe? I just could not make `Job` do what I wanted, with the Moose/Mouse stuff and didn't want to add that axis of confusion.

lib/HTFeed/Stage.pm
mwarin: There is no such test of simultaneous `HTFeed::JobMetrics` writes. 
If you know of an existing similar test I could scavenge, that would be great.

lib/HTFeed/PackageType/Simple/Download.pm
mwarin: No difference beyond length and typeability. 
Switched to long-form name throughout.

lib/HTFeed/Stage/Download.pm
mwarin: Riiight.

I did not think about labels, but it's not a huge change to add.

lib/HTFeed/Stage.pm
mwarin: Class now implements the singleton pattern, changed constructor from `new()` to `get_instance()`.

mwarin: I've added a test where multiple copies of the singleton tries to write at the same time, but having issues (test is green on github, red locally).

Dockerfile
```diff
@@ -60,8 +60,6 @@ RUN echo "deb [signed-by=/usr/share/keyrings/hathitrust-archive-keyring.gpg] htt
 
 RUN apt-get update && apt-get install -y grokj2k-tools
 
-RUN cpan -f -i Net::AMQP::RabbitMQ
```
mwarin, line 63: `Net::AMQP::RabbitMQ` was already included by `Makefile.pl`, which is the proper place to put required modules, so I removed it.

Makefile.PL
```diff
@@ -24,6 +24,7 @@ WriteMakefile(
     'Mouse'               => 0,
     'Net::AMQP::RabbitMQ' => 0,
     'Net::Prometheus'     => 0,
+    'Prometheus::Tiny::Shared' => 0,
```
mwarin, line 27: New required module, used in new class `HTFeed::JobMetrics`.

bin/validate_volume.pl
```diff
@@ -55,35 +55,33 @@
 }
```
mwarin, line 55: Only cosmetic/whitespace changes in this file.

lib/HTFeed/Job.pm
```diff
@@ -1,12 +1,10 @@
 package HTFeed::Job;
```
mwarin, line 1: Mostly cosmetic/whitespace changes in this file.

lib/HTFeed/Stage.pm
mwarin: Now there is a test, AND we're shamelessly green.

lib/HTFeed/Stage/ImageRemediate.pm
mwarin: Added an issue for this. Resolving conversation.

lib/HTFeed/Stage/Pack.pm
mwarin: Made an issue of this. Resolving convo.

lib/HTFeed/Stage/Download.pm
mwarin: Added issue + resolving convo.

lib/HTFeed/JobMetrics.pm
mwarin: added issue(s) + resolving convo

lib/HTFeed/QueueRunner.pm
mwarin: mentioned in an added issue. resolving convo

lib/HTFeed/QueueRunner.pm
mwarin: mentioned in added issue. resolving convo

lib/HTFeed/QueueRunner.pm
```diff
@@ -104,19 +106,34 @@ sub validate_job {
...
-  while($job){
-    get_logger()->info("next job: " . $job->namespace . "." . $job->id . " " . $job->stage_class);
-
-    $job->run_job($self->{clean});
-    $job = $job->successor;
-  }
+    my $self = shift;
+    my $job = shift;
+
+    while($job){
```
mwarin, line 115: A lot of metrics get recorded here.

lib/HTFeed/Stage.pm
```diff
@@ -23,11 +23,14 @@ sub new {
...
     }
 
+    # No class inheriting from HTFeed::Stage needs to instantiate
+    # its own $self->{job_metrics} for storing JobMetrics.
     my $self = {
         volume => undef,
         @_,
         has_run => 0,
         failed  => 0,
+	job_metrics => HTFeed::JobMetrics->get_instance
```
mwarin, line 33: This makes it so that each object that inherits from `Stage` should be able to access `$self->{job_metrics}`, so we don't need to say `use HTFeed::JobMetrics` in every `Stage`-descendant.

lib/HTFeed/Stage/DirectoryMaker.pm
```diff
@@ -1,11 +1,11 @@
 package HTFeed::Stage::DirectoryMaker;
```
mwarin, line 1: Just re-ordering the use-declarations.

aelkiss:
I thought I had re-reviewed this, but apparently the comment didn't show up. The gist was that all the changes looked good, especially the metrics names & units; I still wanted to see the integration tests:

> One thing to address right now is a couple additional tests:
>
> * something involving at least one Stage that verifies that it records the expected metrics (not necessarily the exact values, just some nonzero value)
> * an integration test involving QueueRunner and multiple Stages and again verifies we see non-zero values for expected metrics.

I see you pushed some commits yesterday so I will take a look at those.

aelkiss (approved):
The integration test looks good to me. The way they work with `match` and `get_value` (fishing through the rendered output by regex) is the way I've done integration testing with metrics before, but since we're using those functions exclusively as test helpers (I think) I'd rather see them in `job_metrics.t`. I don't think that needs to block merging, so I'll add an issue for it (#123)

PR #124, opened 2024-05-10
TIDY enqueue

mwarin:
Trying to clean up the enqueue-related code in order to take care of some TODOs (from https://hathitrust.atlassian.net/wiki/spaces/ETT/pages/2928082947/Potential+ingest+work).

Should be mostly whitespace / reordering with no functional changes.

One commit per tidied file for ease of reviewing, that I'll squash post-review / pre-merge.

### DIFF
--- a/lib/HTFeed/Volume.pm
+++ b/lib/HTFeed/Volume.pm
@@ -28,748 +28,825 @@ my $stage_map = undef;
 use constant HT_UUID => '09A5DAD6-3484-11E0-9D45-077BD5215A96';
 
 sub new {
-  my $package = shift;
+    my $package = shift;
 
-  my $self = {
-    objid     => undef,
-    namespace => undef,
-    packagetype => undef,
-    uuidgen => new Data::UUID,
-    @_,
+    my $self = {
+	objid       => undef,
+	namespace   => undef,
+	packagetype => undef,
+	uuidgen     => new Data::UUID,
+	@_,
+	# files	    => [],
+	# dir	    => undef,
+	# mets_name => undef,
+	# mets_xml  => undef,
+    };
 
-    #		files			=> [],
-    #		dir			=> undef,
-    #		mets_name		=> undef,
-    #		mets_xml		=> undef,
-  };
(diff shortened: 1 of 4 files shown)


lib/HTFeed/Volume.pm
```diff
@@ -28,748 +28,825 @@ my $stage_map = undef;
...
-    while( my ($key,$val) = each (%{ $nspkg_filegroups })) {
-      my $files = [];
-      my $re = $val->{file_pattern} or die("Missing pattern for filegroup $key");
-      foreach my $file ( @{ $self->get_all_directory_files() }) {
-        push(@$files,$file) if $file =~ $re;
-      }
-      $filegroups->{$key} = new HTFeed::FileGroup($files,%$val);
+    my $self = shift;
+
+    if(not defined $self->{filegroups}) {
```
moseshll, line 114: An exception to the "space after if" rule that seems to be in place elsewhere. Looks like there are some other examples in this file too, also I'm in favor of "space after while" and I see an instance of a `while( ...` just below. I think it would be worth doing a search on "if(" and "while(" and aligning the whitespace with the overall standard.

lib/HTFeed/Volume.pm
```diff
@@ -28,748 +28,825 @@ my $stage_map = undef;
...
-  }
-  closedir($dh) or croak("Can't closedir $stagedir: $!");
-  @{ $self->{directory_files} } = sort( @{ $self->{directory_files} } );
-#    }
+    my $self = shift;
+    my $path = shift;
+
+    $self->{directory_files} = [];
+    my $stagedir = $path || $self->get_staging_directory();
+    opendir(my $dh,$stagedir) or croak("Can't opendir $stagedir: $!");
```
moseshll, line 138: This seems to be an exception to the "space after param comma" standard, which I also endorse. Ditto my comment about searching for other exceptions, there seem to be some but not a lot.

moseshll (approved):
Two comments re spacing conventions I recommend. I like what I see here. APPROVE.

mwarin:
deleting a merged branch

PR #125, opened 2024-05-23
Use healthcheck instead of wait-for

aelkiss:
Also, remove extra docker-compose.test.yml


aelkiss:
@mwarin I think all the race condition issues we were seeing were in `feed-internal`, but this still seems to work insofar as tests are passing.

docker-compose.yml
```diff
@@ -71,16 +111,25 @@ services:
       MINIO_ACCESS_KEY: TESTINGACCESSKEY
       MINIO_SECRET_KEY: testingsecretkey
     command: server /data
+    healthcheck:
+      <<: *healthcheck-defaults
```
mwarin, line 115: Could minio and pushgateway health checks use the same tool (curl/get) to call the thing they are calling?

mwarin (approved):
Works & looks good. 
Approve!

docker-compose.yml
aelkiss: Sadly, no, because the health check runs in the containers, and those containers are built off of different base images and include different things `prom/pushgateway` is built off of busybox and `minio/minio` appears to be built off of some kind of minimal RHEL thing.

```
aelkiss@byrd:~$ docker run --rm -ti --entrypoint /bin/bash minio/minio
bash-5.1# curl
curl: try 'curl --help' for more information
bash-5.1# wget
bash: wget: command not found
```

```
aelkiss@byrd:~$ docker run --rm -ti --entrypoint /bin/sh prom/pushgateway
/pushgateway $ curl
/bin/sh: curl: not found
/pushgateway $ wget
BusyBox v1.36.1 (2023-12-04 22:44:12 UTC) multi-call binary.

Usage: wget [-cqS] [--spider] [-O FILE] [-o LOGFILE] [--header STR]
...
```

PR #126, opened 2024-05-30
Prometheus exporter for metrics

aelkiss:
This adds a very simple plack application which we can run as a sidecar container for the ingest workers that will allow the metrics to be scraped.

To see it in action/test it locally:

```
docker compose up prometheus-exporter
curl http://localhost:9090 # should see the metric names
cp t/fixtures/simple/test/*.zip volumes_to_test # or put anything else in volumes_to_test
docker compose run validate
curl http://localhost:9090 # should see some values for some metrics
```

### DIFF
--- a/bin/metrics_exporter.pl
+++ b/bin/metrics_exporter.pl
@@ -1,3 +1,6 @@
+use strict;
+use warnings;
+
 use HTFeed::JobMetrics;
 use Plack::Response;
 
@@ -10,7 +13,6 @@ my $app = sub {
   my $env = shift;
 
   my $rendered = $metrics->pretty();
-  my $headers = { "Content-Type" => "text/plain" };
 
   my $response = Plack::Response->new();
   $response->content_type('text/plain');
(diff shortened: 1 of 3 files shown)


aelkiss:
FYI @moseshll

bin/metrics_exporter.pl
```diff
@@ -0,0 +1,30 @@
+use HTFeed::JobMetrics;
```
mwarin, line 4: I would expect to see `use strict;` and `use warnings;` at the top of every perl script, OR a comment with a compelling argument why not.

mwarin (approved):
Looks great. I was able to variations on the following:

```
$ docker compose run --rm prometheus-exporter perl lib/HTFeed/JobMetrics.pm --operation add --metric ingest_handle_items --value 5000

...

$ curl http://localhost:9090 | grep ingest_handle_items
# TYPE ingest_handle_items counter
ingest_handle_items 5000
```

... and that's an APPROVE.

PR #127, opened 2024-06-03
DEV-1120: Metadata fields not copied

mwarin:
## What/why?

Under some circumstances, during validation of submitted items, we lose image metadata that we rely on in later stages of validation.

In this case, we were losing the field `Artist` during image remediation, which is required in a later stage, causing validation to fail.

## What is the fix?

Put Artist back in the metadata. There are also some tidyings. I've tried to group the commits thematically (and will squash them upon approval).

## Testing instructions

Run these commands in order. Look in https://github.com/hathitrust/feed/pull/127/commits to see the commits mentioned.

```
# setup
git clone https://github.com/hathitrust/feed
cd feed
git checkout DEV-1120                                 # Not sure if needed, but dont hurt.
git checkout 26a7bdf21a318b58cf9884ba1a5d97e7e5ae4ec2 # Includes fixture and test, but not the code fix.
docker compose build
docker compose up -d

# part 1: should fail
cp t/fixtures/volumes/bitonal_tiff.zip volumes_to_test/
docker compose run --rm validate
docker compose run --rm test perl t/local_ingest.t

# part 2: should work
git checkout 4905bfcc29209ca44606dd5ed363fc86f0316872 # Includes the code fix.
docker compose run --rm validate
docker compose run --rm test perl t/local_ingest.t
```

To **"fail"** means:

Output from `compose run --rm validate` containing `ERROR: Missing field value	objid: bitonal_tiff	namespace: test	file: 00000001.tif`

Output from `docker compose run --rm test perl t/local_ingest.t` ending with `Looks like you failed 1 test of 19.`

To **"work"** means: 

No such error message in the output.

## Acknowledgements

* I acknowledge that I can't say for sure that this is the best way / place to add the fix.
* I acknowledge that the new test is a bit weak.


lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -430,11 +432,28 @@ sub repair_tiff_imagemagick {
 	"TIFF_REPAIR: attempting to repair $infile to $outfile\n"
     );
 
+    my $in_exif = Image::ExifTool->new;
```
mwarin, line 435: Functional (non-TIDY) changes lines 435-456

lib/HTFeed/PackageType/Simple/ImageRemediate.pm
```diff
@@ -1,84 +1,88 @@
...
-                $self->set_from_meta_yml($meta_yml_field,$set_if_undefined,$tiff_field);
-            }
-
-            # force override resolution if it is provided in meta.yml
-            $self->set_from_meta_yml('bitonal_resolution_dpi',$force_fields,'Resolution');
-
-            return ( $force_fields, $set_if_undefined, $file);
-        }
-    ) if @tiffs;
+    if (@tiffs) {
```
mwarin, line 47: This TIDY puts the IF at the beginning of the block and breaks out the anon sub

t/local_ingest.t
```diff
@@ -162,9 +162,16 @@ describe "HTFeed::PackageType::Simple" => sub {
       my $exiftool = Image::ExifTool->new();
       $exiftool->ExtractInfo("$tmpdirs->{ingest}/lossless_jp2_with_xmp/00000001.jp2");
       is($exiftool->GetValue("XMP-tiff:Make"),"Test scanner make");
-
     };
 
+    it "does not lose artist when compressing a bitonal tiff" => sub {
+	my $volume = unpacked_volume("bitonal_tiff");
```
moseshll, line 168: This tab indentation throws off my editor. My recommendation may apply to other code in the PR but it is especially noticeable here. AFAICT this block should be indented with two tabs but is only one, and IMHO (at the risk of igniting a flame war) there should be no tabs at all, just explicit spaces -- whatever the norm is for HathiTrust Perl, seems to be a toss-up between 2 and 4 spaces.

moseshll (approved):
- It was good to see firsthand an example of tidy + functional changes in a single PR that could be segregated by commit in such a way that they could be untangled. I know we talked about this but I didn't understand what the result should look like, so that's pretty useful. I may talk myself back into doing some tidying in a functional PR if I can plan the commits well enough.
- The walkthrough and the functional changes work fine, with the exception of the Minio healthcheck consistently failing so I commented it out. What up with that? I don't remember that ever happening, fortunately I could work around it and I hope the oddity will go way after a reboot or a few updates on my end. Decided not to chase my tail on that distraction further.
- See inline note on tab characters. Tabs are my personal *bête noire* but maybe I'm behind the times. Approving anyway so we don't get hung up on cosmetics.

mwarin:
> See inline note on tab characters. Tabs are my personal _bête noire_ but maybe I'm behind the times.

This motivated to finally get rid of tabs-as-indentation for perl in my editor. Perhaps `perltidy` next.

mwarin:
Squashed commits.

PR #128, opened 2024-06-10
DEV-1203: Additional ingest performance monitoring (collate/storage)

mwarin:
This PR does the following changes to `JobMetrics`:

* Changed metric naming conventions
* Added labels
* Expanded coverage
* TIDYing of some prominent classes
* Split new class into `JobMetrics` and `JobMetricsCLI`, to simplify main class and not pollute it with CLI stuff. 

In most cases where monitoring is added, I've added some variation on:

```perl
my $start_time = $self->{job_metrics}->time;
# ...
# ... the existing piece of code doing the work that we want to measure
# ...
my $end_time   = $self->{job_metrics}->time;
my $delta_time = $end_time - $start_time;
my $labels = {foo => 'bar'}; # something about the operation being measured
$self->{job_metrics}->add("ingest_xxx_bytes_r_total", -s $infile, $labels); # if we did any r/w op
$self->{job_metrics}->add("ingest_xxx_bytes_w_total", -s $outfile, $labels); # if we did any r/w op
$self->{job_metrics}->add("ingest_xxx_seconds_total", $delta_time, $labels); # the time it took
$self->{job_metrics}->inc("ingest_xxx_items_total", $labels); # if we did anything with an item
$self->{job_metrics}->inc("ingest_xxx_images_total", $labels); # if we did anything with an image
```
... plus a matching metric declaration in `HTFeed::JobMetrics`.

I've also added 2 convenience subs to `HTFeed::JobMetrics`:  

* `time` for providing high resolution time  
* `dir_size`  for calculating the size of a directory 

... which are both data points that are often needed for measuring.

## Reviewer info

I originally tried to separate and label commits thematically, but this started falling apart as the PR grew too large. There are some TIDY commits, a big FUNC commit with mostly additions (that nevertheless has some tidyings in it, discipline slipped), and a TEST commit. I will squash them all before merging.

Please check: 
* the new tests in `t/job_metrics.t`
* the changes to `JobMetrics::_setup_metrics`
* that the TIDY commits aren't going too far
* that I'm  measuring what I say I'm measuring (i.e. not using the wrong metric in the wrong place)

Please be on the lookout for any log warnings about `WARN: invalid metric name "ingest_xxx_yyy"`, this is a sign that we are trying to record data for a metric that has not been declared in `HTFeed::JobMetrics`.

### DIFF
--- a/lib/HTFeed/JobMetrics.pm
+++ b/lib/HTFeed/JobMetrics.pm
@@ -2,10 +2,16 @@ package HTFeed::JobMetrics;
 
 use strict;
 use warnings;
-use Getopt::Long;
+use File::Find;
 use Log::Log4perl qw(get_logger);
-use Pod::Usage;
 use Prometheus::Tiny::Shared;
+use Time::HiRes;
+
+# TODO: mw 2024 // simplify predefined metrics.
+# Since it is already starting to grow out of hand...
+# I still want metric base names to be predefined, and a predefined set
+# of units. E.g. @metrics = qw(ingest_download ingest_pack)
+# and @units = qw(seconds bytes_r bytes_w items images)
 
 =item HTFeed::JobMetrics
 
@@ -16,11 +22,10 @@ the various stages of an ingest job, such as number of items downloaded,
 number of bytes downloaded and amount of time spent downloading
 (as well as the other job stages, not just download).
 
-This class implements the singleton pattern, and as such has a
-get_instance() method instead of a "normal" new() function.
+This class implements the singleton pattern. The singleton instance is
+stored in the aptly named $singleton variable.
 
-The singleton instance is stored in the aptly named
-$singleton variable.
+CLI provided by and documented in JobMetricsCLI.pm.
 
 =item See e.g.:
 
@@ -28,253 +33,131 @@ $singleton variable.
 
 =back
 
-=item Command line options:
-
-=back
-
-=item --help / --usage
-
-Display the POD snippets in this file as a help document.
-
 =cut
-
-# If called from terminal:
-if (!caller) {
-    GetOptions(
-	# these are all booleans
-	'clear'    => \my $clear,
-	'help'     => \my $help,
-	'list'     => \my $list,
-	'location' => \my $location,
-	'pretty'   => \my $pretty,
-	'usage'    => \my $usage,
-	# Each operation takes a metric and/or a value:
-	# add(metric, value),
-	# get_value(metric)
-	# inc(metric),
-	# match(value),
-	# observe(metric, value)
-	'operation=s' => \my $operation,
-	'metric=s'    => \my $metric,
-	'value=s'     => \my $value
-    );
-
-    if ($help or $usage) {
-	pod2usage(2);
-    }
-
-    my $job_metrics = HTFeed::JobMetrics->get_instance();
-    my %valid_operations = (
-	add       => sub { $job_metrics->add($metric, $value) },
-	inc       => sub { $job_metrics->inc($metric) },
-	match     => sub {
-	    my $matches = $job_metrics->match($value);
-	    print join("\n", @$matches) . "\n";
-	},
-	get_value => sub {
-	    my $value = $job_metrics->get_value($metric);
-	    if (defined $value) {
-		print "$value\n";
-	    } else {
-		print "no value found\n";
-	    }
-	},
-	observe => sub { $job_metrics->observe($metric, $value) }
-    );
-
-    # Check the boolean args and act if true.
-    if ($clear) {
-	print "Clearing data...\n";
-	$job_metrics->clear;
-	print "Data cleared.\n";
-    }
-    if ($list) {
-	print "# List of metrics:\n";
-	print join("\n", @{$job_metrics->list_metrics}) . "\n";
-    }
-    if ($location) {
-	print "# HTFeed::JobMetrics data stored in:\n";
-	print $job_metrics->loc . "\n";
-    }
-    if ($pretty) {
-	print $job_metrics->pretty . "\n";
-    }
-
-    # Now check the operation arg
-    if (defined $operation) {
-	if (exists $valid_operations{$operation}) {
-	    print "running operation $operation ...\n";
-	    $valid_operations{$operation}();
-	} else {
-	    print "Invalid operation, exiting.\n";
-	    exit(1);
-	}
-    } else {
-	print "No operation defined, exiting.\n";
-	exit(0);
-    }
-}
-# end commandline stuff
 
 my $singleton = undef;
 
-sub get_instance {
+# This class implements the singleton pattern (but constructor is still new()).
+sub new {
     # Re-use singleton if defined.
     if (!defined $singleton) {
-	my $class = shift;
-	my $self  = {};
-	# The default location for storing metrics data is under /tmp/.
-	# You can ovverride that w/ $ENV{'HTFEED_JOBMETRICS_DATA_DIR'}.
-	$self->{file} = "/tmp/htfeed-jobmetrics-data";
-	if (defined $ENV{'HTFEED_JOBMETRICS_DATA_DIR'}) {
-	    $self->{file} = $ENV{'HTFEED_JOBMETRICS_DATA_DIR'};
-	}
-	$self->{prom} = Prometheus::Tiny::Shared->new(
-	    filename => $self->{file}
-	);
-	$singleton = bless($self, $class);
-	$singleton->_setup_metrics();
+        my $class = shift;
+        my $self  = {};
+        # The default location for storing metrics data is under /tmp/.
+        # You can ovverride that w/ $ENV{'HTFEED_JOBMETRICS_DATA_DIR'}.
+        $self->{file} = "/tmp/htfeed-jobmetrics-data";
+        if (defined $ENV{'HTFEED_JOBMETRICS_DATA_DIR'}) {
+            $self->{file} = $ENV{'HTFEED_JOBMETRICS_DATA_DIR'};
+        }
+        $self->{prom} = Prometheus::Tiny::Shared->new(
+            filename => $self->{file}
+        );
+        $singleton = bless($self, $class);
+        $singleton->_setup_metrics();
     }
+
     return $singleton;
 }
 
-=item --loc
-
-Show where the data is stored.
-
-=cut
-
+# Show where the data is stored.
 sub loc {
     my $self = shift;
+
     $self->{file};
 }
 
-=item --list
-
-Show names of all known metrics.
-
-=cut
-
+# Show names of all known metrics.
 sub list_metrics {
     my $self = shift;
+
     [sort keys %{$self->{metrics}}];
 }
 
-=item --pretty
 
-Full, somewhat readable, accounting of known metrics and their values.
-
-Call this to export/harvest/scrape data from the running production pod.
-
-=cut
-
+# Full, somewhat readable, accounting of known metrics and their values.
 sub pretty {
     my $self = shift;
+
     $self->{prom}->format;
 }
 
-=item --clear
-
-Irrevocably delete job metrics (see --loc for where that is)
-
-=cut
-
+# Irrevocably delete job metrics
 sub clear {
     my $self = shift;
+
     $self->{prom}->clear;
 }
 
-=item --operation
-
---operation takes an operation name (match / inc / add/ observe),
-a --metric and/or a --value. See below for the different operations.
-
-=item --operation match --value x
-
-Return metrics matching x.
-
-=cut
-
+# Return matching lines from pretty output
 sub match {
     my $self   = shift;
     my $value  = shift;
     die "match requires a value" if !defined $value;
 
     my @pretties = split("\n", $self->pretty);
+
     [grep { /$value/ } @pretties];
 }
 
-=item --operation get_value --metric x
-
-Returns the value for metric x, 0 in case of any issues.
-
-=cut
-
+# Return value for a given metric
 sub get_value {
     my $self   = shift;
     my $metric = shift;
 
     if (!$self->_valid_metric($metric)) {
-	return 0;
+        return 0;
     }
 
     my $match = $self->match("^$metric");
     if (scalar @$match == 1) {
-	if ($$match[0] =~ /(\d+(?:\.\d+)?)$/) {
-	    my $matched_numerical_value = $1;
-	    return $matched_numerical_value;
-	}
+        if ($$match[0] =~ /(\d+(?:\.\d+)?)$/) {
+            my $matched_numerical_value = $1;
+            return $matched_numerical_value;
+        }
     }
     get_logger()->warn("JobMetric found no value for $metric");
+
     return 0;
 }
 
-=item --operation inc --metric x
-
-Increment metric x by 1.
-
-=cut
-
+# Increment the given metric (w/ optional labels href)
 sub inc {
     my $self   = shift;
     my $metric = shift;
+    my $labels = shift || {};
+
+    # sets to {} if invalid
+    $labels = $self->_valid_labels($labels);
+
     if ($self->_valid_metric($metric)) {
-	$self->{prom}->inc($metric);
+        $self->{prom}->inc($metric, $labels);
     }
 }
 
-=item --operation add --metric x --value y
-
-Adds arbitrary numeric value y to the metric x.
-
-=cut
-
+# Add to the value of the given counter metric (w/ optional labels href)
 sub add {
     my $self   = shift;
     my $metric = shift;
     my $value  = shift;
+    my $labels = shift || {};
+
+    # sets to {} if invalid
+    $labels = $self->_valid_labels($labels);
 
     # Make sure $value is defined and numeric
     unless (defined $value) {
-	get_logger()->warn("Undefined value given for $metric");
-	return;
+        get_logger()->warn("Undefined value given for $metric");
+        return;
     }
-    unless ($value + 0) {
-	get_logger()->warn("Non-numeric value \"$value\" given for $metric");
-	return;
+    unless ($value + 0 >= 0) {
+        get_logger()->warn("Non-numeric value \"$value\" given for $metric");
+        return;
     }
 
-    $self->_valid_metric($metric) && $self->{prom}->add($metric, $value);
+    $self->_valid_metric($metric) && $self->{prom}->add($metric, $value, $labels);
 }
 
-=item --operation observe --metric x --value y
-
-Add value y to the histogram metric x.
-
-Note that histograms must be set up with buckets ahead of time.
-
-=cut
-
+# Add to the buckets of the given bucket metric
 sub observe {
     my $self   = shift;
     my $metric = shift;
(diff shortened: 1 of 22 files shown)


lib/HTFeed/Stage.pm
```diff
@@ -48,14 +48,6 @@ sub estimate_space{
     return ceil($size * $multiplier);
 }
 
-sub dir_size{
```
mwarin, line 51: This sub was moved to HTFeed::JobMetrics.

lib/HTFeed/JobMetrics.pm
```diff
@@ -318,61 +360,74 @@ sub _setup_metrics {
...
+        'stage',
+        'unpack',
+        'validate_zip_completeness',
+        'verify_crypt',
+        'verifymanifest',
+        'volumevalidator',
+    );
+    my @units = (
+        'bytes_read',
+        'bytes_write',
```
moseshll: This naming convention is difficult. I would propose the alternative "bytes_read" and "bytes_written" although it seems ungainly. Alternatives to that: [rw]bytes, bytes[rw], bytes_[in,out], [input,output]_bytes. The main thing I am looking for is for the last element of the metric to be the "thing measured", just trying to work around the fact that two irregular verbs are not fun to contend with.

t/job_metrics.t
```diff
@@ -6,14 +6,15 @@ use FindBin;
...
 
 describe "HTFeed::JobMetrics" => sub {
     # SETUP start
     my $jm = HTFeed::JobMetrics->get_instance;
     my $item_metric = "ingest_pack_items";
-    my $byte_metric = "ingest_pack_bytes";
+    my $byte_metric = "ingest_pack_bytes_write";
     my $invalid_metric = "not a valid metric";
     before each => sub {
 	$jm->clear() if defined $jm;
```
moseshll: I think this line is the first example of some tabs pre-existing/remaining in the .t file. I'd like to see an additional TIDY on this to clear out the rest of the tabs.

moseshll (approved):
See two comments, one on naming conventions and the other on tabs. Don't feel like making either an impediment to merging, though I think the naming convention is worth further thought/brainstorming.

lib/HTFeed/JobMetrics.pm
aelkiss: Looking at existing metrics it should probably end in `bytes_total` or maybe just `bytes` is fine. A couple examples I see from existing metrics:

* container_fs_reads_bytes_total
* container_fs_writes_bytes_total
* node_disk_read_bytes_total
* node_disk_written_bytes_total
* ceph_osd_op_r_out_bytes
* ceph_osd_op_w_in_bytes

I believe the `_total` is to indicate that it's a counter rather than a gauge (e.g. to differentiate it from something like `container_fs_usage_bytes`) -- but the ceph metrics are also counters and don't end in `_total` so 🤷

aelkiss:
@mwarin Have we looked at perltidy https://metacpan.org/pod/Perl::Tidy recently? It might be able to do some of these tidyings/reformattings automatically, although it looks like it has quite a few options...

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -371,6 +374,13 @@ sub _remediate_tiff {
 	$self->{newFields}
     );
 
+    my $end_time   = $self->{job_metrics}->time;
+    my $delta_time = $end_time - $start_time;
+    $self->{job_metrics}->add("ingest_imageremediate_seconds", $delta_time);
+    $self->{job_metrics}->inc("ingest_imageremediate_images");
+    $self->{job_metrics}->inc("ingest_imageremediate_bytes_read", $infile_size);
```
aelkiss: I think you want `add` rather than `inc` here - `inc` always increments by 1.

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -412,21 +425,31 @@ sub repair_tiff_exiftool {
...
+    my $write_return = $exifTool->WriteInfo(@file_params);
+    if (!$write_return) {
+        croak(
+            "Couldn't remediate $infile: ". $exifTool->GetValue('Error') . "\n"
+        );
         return 0;
     }
+    my $end_time = $self->{job_metrics}->time;
+    my $delta_time = $end_time - $start_time;
+    $self->{job_metrics}->inc("ingest_imageremediate_images");
```
aelkiss: We should add some labels here to differentiate between exiftool, imagemagick, tiff repair, etc. It might also make sense to extract this as a subroutine since we are doing something with a very similar shape in multiple places in `ImageRemediate`.

lib/HTFeed/Storage.pm
```diff
@@ -300,26 +331,37 @@ sub move {
...
+        my $bytes_moved = (-s $mets_stage || 0) + (-s $zip_stage || 0);
+        $self->safe_system('mv', '-f', $mets_stage, $object_path);
+        $self->safe_system('mv', '-f', $zip_stage,  $object_path);
+
+        get_logger->debug("  finished move");
+        my $end_time    = $self->{job_metrics}->time;
+        my $delta_time  = $end_time - $start_time;
+        $self->{job_metrics}->inc("ingest_move_items");
+        $self->{job_metrics}->add("ingest_move_seconds", $delta_time);
+        $self->{job_metrics}->add("ingest_move_bytes_read", $bytes_moved);
```
aelkiss: This may not be super relevant since the staging path should typically be on the same filesystem as the object path, so the move becomes an atomic rename rather than a copy that requires reading/writing the items again..

lib/HTFeed/Storage.pm
```diff
@@ -51,17 +52,18 @@ sub new {
...
+        volume         => $volume,
+        namespace      => $volume->get_namespace(),
+        objid          => $volume->get_objid(),
+        errors         => [],
+        mets_source    => $volume->get_mets_path(),
+        zip_source     => $volume->get_zip_path(),
+        zip_suffix     => '',
+        config         => $config,
+        did_encryption => 0,
+        name           => $name,
```
aelkiss, line 64: We should record the name of the storage as a label for all metrics recorded here.

lib/HTFeed/Storage.pm
aelkiss: We should also review the subclasses (especially `HTFeed::Storage::ObjectStore` since there are some different implementations that may override the metrics recording here.

aelkiss (commented):
Overall I think this is looking like a good approach. There are a couple things I'd like to see:
* labels for the storage name
* labels for the specific image remediate operation
* consider extracting the metrics recording stuff to a method to reduce duplication

t/job_metrics.t
```diff
@@ -95,7 +96,50 @@ describe "HTFeed::JobMetrics" => sub {
...
+        my $t1                     = $jm->time;
+        my $sleep_time             = 0.1;
+        my $sleep_time_in_microsec = $sleep_time * 1000000;
+        my $resolution_tolerance   = 0.01;
 
+        usleep $sleep_time_in_microsec;
+        my $t2      = $jm->time;
+        my $delta_t = $t2 - $t1;
+
+        # Now prove that t1 and t2 are (very close to) 0.25 sec apart.
```
aelkiss: I think the logic here is correct, but the actual number doesn't match the `$sleep_time` listed above

t/job_metrics.t
mwarin: Re-indented: https://github.com/hathitrust/feed/pull/128/commits/79b0b9577fb2217d8ade59a20a46a680c001b86d?w=0

lib/HTFeed/Stage/ImageRemediate.pm
mwarin: Yep. Good catch. All other hits when grepping for `'inc\(.+s'` are for items and images, as they should.

t/job_metrics.t
mwarin: corrected

lib/HTFeed/Storage.pm
mwarin: I've added labels to `HTFeed::JobMetrics`, so it is possible to record them. Now a matter of implementing where desired.

lib/HTFeed/Stage/ImageRemediate.pm
mwarin: @aelkiss maybe it is a shortcoming of my imagination, but any extracted subroutine I picture is going to be about as complicated to call as these existing lines. Each time we make a (set of) measurement(s) of an operation we need to determine 1) when the operation started, 2) ended, 3) which file sizes meaningfully changed, 4) which labels to use, 5) which metric to increment, 6) ...?

I can imagine writing the time measurement as a callback:

```
my $timer = $job_metrics->start_timer("ingest_pack_items", $labels);
... # time passes, work happens
$timer->stop_and_record(); # writes delta time since get_timer to prometheus
```
... but how would I do that for file & dir sizes? I guess I could pass in the paths for inputs and outputs. But at this point I'm using about as many lines of code each time setting up callbacks as I'm currently using, and I'm starting to lose the point.

Do tell me if you have strong feelings about this, or if I'm misunderstanding your feedback.

aelkiss: I think fine to leave as-is for now. I'll take a look and see if I can propose something at a future point - https://github.com/hathitrust/feed/issues/134

lib/HTFeed/QueueRunner.pm
```diff
@@ -121,22 +121,7 @@ sub run_job_sequence {
...
-        # Get the verby part of the stage class name (e.g. "pack")
-        # that we need for job metrics.
-        my $jobtype    = lc((split("::", $job->stage_class))[-1]);
-        my $start_time = $self->{job_metrics}->time;
-
         $job->run_job($self->{clean});
-
-        # TODO: it's easy to count seconds and items for jobs out here
-        # but we have to go inside the job classes to count bytes r/w.
-        my $end_time       = $self->{job_metrics}->time;
```
mwarin: This was how and where a lot of metrics used to be set, "for free" by `QueueRunner` as it runs each stage, but that does not work if we want labels or measuring things at a scale smaller than stages (e.g. individual `ImageRemediate` operations). For that we need to go inside each stage and look closer at the inputs/outputs. So this piece of code sadly has to go, or we'll run the risk of counting things more than once.

lib/HTFeed/Storage.pm
mwarin: No longer measuring `sub move`: https://github.com/hathitrust/feed/pull/128/commits/7f35bc7cfe702d0a168ac0b9be4f01f2b1a8b55e

lib/HTFeed/Storage.pm
mwarin: Added storage name as a label value:https://github.com/hathitrust/feed/pull/128/commits/7f35bc7cfe702d0a168ac0b9be4f01f2b1a8b55e

lib/HTFeed/JobMetrics.pm
mwarin: I'm leaning towards `_r_` and `_w_` as these strings are silly long already. And speaking of long, I'm adding `_total` to all current metric names (because they are all counters).

lib/HTFeed/JobMetricsCLI.pm
```diff
@@ -0,0 +1,178 @@
+package HTFeed::JobMetricsCLI;
```
mwarin, line 1: This whole class was broken out of `HTFeed::JobMetrics`.

lib/HTFeed/Storage.pm
mwarin: Made issue #136 for `HTFeed::Storage::ObjectStore`.

lib/HTFeed/JobMetrics.pm
```diff
@@ -22,273 +28,178 @@ get_instance() method instead of a "normal" new() function.
...
-	print "No operation defined, exiting.\n";
-	exit(0);
-    }
-}
-# end commandline stuff
-
 my $singleton = undef;
 
+# This class implements the singleton pattern.
+# Call get_instance instead of new.
```
aelkiss: @moseshll had preferred `new` on the imgsrv side; we discussed a bit and he convinced me (https://github.com/hathitrust/babel/pull/13#discussion_r1664646616). Doesn't need to block merging, probably worth an issue to change it later (https://github.com/hathitrust/feed/issues/142)

lib/HTFeed/JobMetrics.pm
```diff
@@ -308,73 +219,123 @@ sub _valid_metric {
...
     # die ("invalid metric name \"$metric\" at $filename:$line");
+
     return 0;
 }
 
+# Labels must be sent as a hashref with simple key-value pairs.
+# Not going to impose any stricter validation at this point.
+# Keys and values will be lowercased.
+# Labels do NOT need to be predefined, unlike metrics.
+sub _valid_labels {
```
aelkiss, line 226: What happens if we put invalid labels in the output? Does `Prometheus::Tiny::Shared` refuse to use them / raise an error? Or does it put garbage in its output and potentially cause issues down the line of scraping/using those metrics? If the latter, this is definitely worth having. If the former, what extra does this validation routine add?

lib/HTFeed/JobMetrics.pm
```diff
@@ -308,73 +219,123 @@ sub _valid_metric {
...
-	'verifymanifest',
-	'volumevalidator'
+        'collate',
+        'download',
+        'download_dropbox',
+        'download_google',
+        'download_ia',
+        'handle',
+        'imageremediate',
+        'encrypt',
```
aelkiss, line 275: I think if we find ourselves having to change this much more in the future we may want to rethink enumerating these here. I'm not sure how to track that sort of "hypothetical issue"... maybe that's a good place for a note in the code?

lib/HTFeed/JobMetricsCLI.pm
aelkiss: I think that helps separate responsibilities well.

lib/HTFeed/Stage/Download.pm
```diff
@@ -63,7 +63,7 @@ sub download{
 
     if( $response->is_success() ){
         my $size = (-s $pathname);
-	$self->{job_metrics}->add("ingest_download_bytes", $size);
+	$self->{job_metrics}->add("ingest_download_bytes_r_total", $size);
```
aelkiss, line 66: I forget why the `r` / `w` thing came up; I don't think it's necessary here insofar as download implies a certain thing (both reading bytes over the network and writing them out to disk); doesn't need to block merge.

lib/HTFeed/Stage/Pack.pm
```diff
@@ -104,11 +104,13 @@ sub zip{
...
-    # Get size after zipping and report size delta as a job metric for Pack.
-    my $end_size = -s $zip_path;
-    my $delta_size = $start_size - $end_size;
-    $self->{job_metrics}->add("ingest_pack_bytes", $delta_size);
-
+    my $end_time = $self->{job_metrics}->time;
+    my $delta_time = $end_time - $start_time;
+    $self->{job_metrics}->add("ingest_pack_seconds_total", $delta_time);
+    my $bytes_w = -s $zip_path;
+    $self->{job_metrics}->add("ingest_pack_bytes_r_total", $bytes_r);
```
aelkiss, line 111: Here we have `r` vs. `w` that is the minimal difference between the metrics so definitely does help differentiate/clarify

lib/HTFeed/Stage/Unpack.pm
```diff
@@ -87,10 +92,15 @@ sub _extract_file {
         return;
     }
 
+    # We don't use $self here, so we have to get job_metrics a different way.
+    my $end_time    = $job_metrics->time;
+    my $delta_time  = $end_time - $start_time;
+    $job_metrics->add("ingest_unpack_seconds_total", $delta_time);
+    my $outdir_size = $job_metrics->dir_size($outdir);
+    $job_metrics->add("ingest_unpack_bytes_w_total", $outdir_size);
```
aelkiss, line 103: unpack is also reading from something to unpack; maybe we should record that as well? That can be a future issue for sure.

aelkiss (approved):
This is looking in pretty good shape; I don't see anything I think needs to block merge, although there are a couple small things to potentially consider now and a few things I think we should note as issues for future work.

I did want to verify that although this PR adds support for labels on metrics, there aren't yet places we're using those labels -- correct?

lib/HTFeed/JobMetrics.pm
mwarin: I don't care either way. I'm happy to change back to `new` for the sake of consistency.

mwarin: fixed in [de8df11](https://github.com/hathitrust/feed/pull/128/commits/de8df11fdd0fab74b8cbfb0632ea81c8d8aa0e7c)

lib/HTFeed/JobMetrics.pm
mwarin: It's mostly training wheels, so I don't start implementing metric names with typoes in them. Prometheus is pretty chill about metric names. I don't trust myself to read, write, or even copy/paste e.g. `validate_zip_completeness_bytes_r_total` consistently, and if you spell it differently in 2 classes you're going to get your data split into 2 measurements.

But I can see that these training wheels are becoming a liability and a limiting factor going forwards.

mwarin: Prometheus is silently discarding any labels it doesn't like, which is why I added this validation.

lib/HTFeed/Stage/Download.pm
mwarin: I thought you wanted me to specify _r/_w in general, and then it made sense to stick with _r even when implied, for consistency.

lib/HTFeed/Stage/Unpack.pm
mwarin: Yep, missed it. Added in [46341c8](https://github.com/hathitrust/feed/pull/128/commits/46341c80af90d93ad08ab9da5be4e71748c5e41b)

lib/HTFeed/JobMetrics.pm
mwarin: I wanted/needed some structure to keep myself consistent as I was implementing metric names and labels. Totally free-form metric names would be near-useless, and metric names with typoes would be frustrating. But I agree that this is not sustainable and should be scaled back as we settle into use. (For instance, my early insistence that metrics must start with "ingest_" is already limiting/silly).

I think ultimately we can be satisfied with `/^\w+(_\w+)*$/` (`word1_word2_..._wordn`).

mwarin: And in case I didn't express this already, the main reason for this is because Prometheus requires that metrics be declared before they can be used. Perhaps we break this out as `JobMetrics::Definitions` which would be mostly lists of metric names and a few helper methods ...
```perl
package HTFeed::JobMetrics::Definitions;
my @counter_metrics = qw( ... );
foreach my $metric (@counter_metrics) {
    $prom->declare($metric_name, type => "counter");
} 
```

Flatter, dumber, easier to add/edit.

mwarin: Added issue #143 for this.

lib/HTFeed/Stage/Download.pm
aelkiss: Right, I think just for download it's kind of dual purpose (reading from the network, writing to disk), but it's not a big deal.

lib/HTFeed/JobMetrics.pm
aelkiss: Right, I don't have a good sense of what the preferred pattern should ultimately be for registering metrics, and it might be worth coordinating with whatever we do on the babel side. I think insisting on `ingest_` as the prefix here is actually good because prometheus doesn't have any kind of namespacing. In any case this works for now and we can comment further on #143.

PR #131, opened 2024-06-25
DEV-1239: pinned RabbitMQ to the last working version

mwarin:
Originating ticket: https://hathitrust.atlassian.net/browse/DEV-1239

This is a temporary solution.
The last working version is 2.40010.

In the long run, we need to figure out how to update dependencies for the latest version (2.40011) to work, but that's a separate task.

Reviewer: Verify that the docker image builds and tests run.


kron-spar (approved):
Can see that the version we want is pinned directly in the make file. Looks good.

PR #132, opened 2024-06-26
Updated the health check for Minio

kron-spar:
I updated the docker-compose.yml file and built the services. The service came online without an issue.


mwarin (approved):
This builds locally for me & tests pass. I can see the same in the github action that ran cloud-side.

I'm also happy to see the `mc` solution, which feels way more proper than the other potential solution I had stumbled across. I like the idea of containers taking responsibility for providing their own health checks.

Issue #135, opened 2024-07-02
Remove `Net::Prometheus` as a dependency


aelkiss:
It looks like we're using both `Net::Prometheus` and `Prometheus::Tiny::Shared` but only using the latter

mwarin:
I'll remove it as part of https://github.com/hathitrust/feed/pull/141 (a mostly deletes PR anyways)

PR #141, opened 2024-07-23
Removing (as of late) unused PackageTypes from feed

mwarin:
Part of DEV-1190

Bunch of deleted files, guided by which packagetypes have been used in the last years & input from Aaron.

In addition, a few changes to the TEST namespace (to stop using removed packagetypes) and removing the dependency to the unused `Net::Prometheus` (https://github.com/hathitrust/feed/issues/135).


aelkiss:
I would also recommend removing:
* Kirtas
* MDLContone and its ilk
* Yale

The MDLContone stuff was an old pilot; there are things in the repository but I don't think we need support for ingesting them. The Kirtas and Yale material was old one-offs and if we did something like that in the future we would likely first want to transform it to the 'simple' package type.

We also don't currently have any epub packages in the repository or any plans to ingest them, but that code is at least tested. I waffle about removing it. I suppose if/when we did epub we could bring it back. I'm just not sure. What do you think?

I think we could also remove `WAVE_hul.pm` from `ModuleValidator`, which supports the audio stuff.

mwarin:
@aelkiss if epub stuff is somewhat under test then we'd be throwing out a lot of baby along with the bathwater. 
I say leave it in for now. We're doing plenty of pruning as it is.

mwarin:
@aelkiss could you please check if I missed anything?

aelkiss (approved):

PR #145, opened 2024-07-26
TIDY ImageRemediate

mwarin:
Mostly whitespace changes, so recommend viewing diff with:
https://github.com/hathitrust/feed/pull/145/files?diff=unified&w=1

The occasional other change is in the form of harmonizing the way we get input args to subs:

```perl
sub abc {
    my $self = shift;
    my ($other, $params, $like, $this) = @_;
    # ...
}
```
... is functionally equivalent to, perhaps more compact than, but not as good looking as:

```perl
sub abc {
    my $self   = shift;
    my $other  = shift;
    my $params = shift;
    my $like   = shift;
    my $this   = shift;
    # ...
}
```

... or in the way default values to arguments are done.

```diff
-    my $force_headers            = ( shift or {} );
+    my $force_headers            = shift || {};
```


carylwyatt (approved):
Looks very tidy!

I had to look up `shift`. I learned something new today, thanks!

Approved ✅

PR #146, opened 2024-07-31
Classes to encapsulate Grok & ImageMagick system calls

mwarin:
This PR mainly moves some code out of inline `system()` calls and into more "proper" perl classes. There isn't much (any?) functional change... we're still doing the same things, just using different code to do it.

* Tidied a bit
* Broke out system calls from `HTFeed::Stage::ImageRemediate` into new classes `HTFeed::Image::Grok` and `HTFeed::Image::Magick` (plus some shared functionality in `HTFeed::Image::Shared`)
* Added tests for new classes
* Added new dev tool / test util: `fail_fast.sh`
* Added a directory for reference images, thinking this will make it easier to test image operations

---

## Review request

I'd like eyeballs on the 2 new classes `Grok` & `Magick` & new test `image.t`.
Check that you can build & test green:

```
git clone https://github.com/hathitrust/feed.git
cd feed
docker compose build
docker compose run --rm test bash t/fail_fast.sh t/*.t
```

The fail_fast.sh output should, if all was OK, end with:
```
volume.t : t/volume.t
```


mwarin:
I currently can't figure out why `t/local_ingest.t` fails. I think I'm using the same exact flags, but the md5sum of the generated file is different (but consistent) when comparing dev branch to main:

```
/usr/bin/convert -compress Group4 '.../feed-test-preingest-X/bitonal_tiff/00000001.tif' -strip '.../feed-test-ingest-Y/bitonal_tiff/00000001.tif'
```

liseli (commented):
Executing this command `docker compose run --rm test bash t/fail_fast.sh t/*.t`

I see the error below. Although it seems you did not update the t/classtests.t script, this test could be impacted with four changes.

`lisepul@2379591497 feed % docker compose run --rm test bash t/fail_fast.sh t/*.t
[+] Creating 4/0
 ✔ Container feed-rabbitmq-1     Running                                                                                                       0.0s 
 ✔ Container feed-pushgateway-1  Running                                                                                                       0.0s 
 ✔ Container feed-minio-1        Running                                                                                                       0.0s 
 ✔ Container feed-mariadb-1      Running                                                                                                       0.0s 
001_load.t : t/001_load.t
002_load_all.t : t/002_load_all.t
backup_expiration.t : t/backup_expiration.t
gpg: error opening lockfile '/usr/local/feed/.gnupg/pubring.kbx.lock': No such file or directory
gpg: lockfile disappeared
bunnies.t : t/bunnies.t
classtests.t : t/classtests.t
HTFeed::Namespace::Test is not Test::Class or integer at t/classtests.t line 26.
Error at t/classtests.t
see /tmp/perltest_results/classtests.t
`

mwarin:
@liseli I don't understand why that error would happen, and you're right that's not part of the changed files. 
When confronted by mysterious errors I usually blame the cache, nuke the cache, and run the tests again. 
You could also run the tests in the same way as .github/workflows/tests.yml :

```
docker compose run test-and-cover
```

liseli (approved):
Due to some unexpected issues, I could not run the tests locally. Martin and I set together to try to fix but it seems so complex. This PR does not introduce new features, it only gets a better organization of the functionalities. The change makes sense to me and I will approve it.

PR #147, opened 2024-08-02
Dev 1253 whitebalance

mwarin:
Now when `ImageRemediate` sees a pagescan with `WhiteBalance=4` it'll try and be OK with it, rather than bail out of processing the item.

Added to `@imagemagick_remediable_errs` :
```
'WhiteBalance value out of range: 4',
```

In order to test it locally, I added yet another version of `validate_volume` : 

```
# Call from outside container
$ docker compose run --rm test bash bin/validate_volume.sh volumes_to_test/volume.zip

# Call from inside container
$ docker compose run --rm test bash 
ingest@bb3a99453802:~$ bash bin/validate_volume.sh volumes_to_test/volume.zip
```
I don't know what's up with the other `validate_volume` scripts, but I couldn't get them working.
I put together a test item at `/htprep/failed/test/whitebalance4.zip` rather than adding it to the repo.

### DIFF
--- a/bin/validate_volume.sh
+++ b/bin/validate_volume.sh
@@ -1,0 +1,12 @@
+# bash validate_volume.sh /path/to/item.zip
+
+fullpath=$(realpath $1)
+objid=$(basename $fullpath .zip)
+test_dir=/tmp/prep/toingest/test/
+
+# Set up working dir and copy input file there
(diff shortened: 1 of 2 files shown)


aelkiss (approved):
Works for me. I appreciate the `validate_volume` wrapper.

PR #148, opened 2024-08-05
DEV-1313 (Local ingest doesn't look for material in "failure" directory)

mwarin:
`bin/enqueue.pl` is an important script for doing operational ingest work, but has not had any tests and is (according to the originating ticket) sometimes behaving oddly. I have been unable to replicate the error condition as described in the originating ticket, but: 

* tidied `t/queue.t`
* added `t/enqueue.t` (for testing `bin/enqueue.pl`)

Having `t/enqueue.t` will make it easier to test the error condition if/when I ever get closer to it.

### DIFF
--- a/t/enqueue.t
+++ b/t/enqueue.t
@@ -1,0 +1,186 @@
+use strict;
+use warnings;
+
+use FindBin;
+use lib "$FindBin::Bin/lib";
+
+use HTFeed::Config qw(get_config);
+use HTFeed::DBTools qw(get_dbh);
+use HTFeed::Queue;
+use HTFeed::Volume;
+use Test::Exception;
+use Test::Spec;
+
+describe "bin/enqueue.pl" => sub {
+
+    sub testvolume {
+	HTFeed::Volume->new(
+	    packagetype => 'simple',
+	    namespace   => 'test',
+	    objid       => 'test'
+	);
+    }
+
+    sub fake_bibdata {
+	my $volume = shift;
+	my $sql = join(
+	    " ",
+	    "REPLACE INTO feed_zephir_items",
+	    "(namespace, id, collection, digitization_source, returned)",
+	    "VALUES",
+	    "(?,         ?,  'TEST',     'test',              '0')"
+	);
+	get_dbh()->do($sql, {}, $volume->get_namespace, $volume->get_objid);
+    }
+
+    sub enqueue_testvolume {
+	my %params = (
+	    # defaults
+	    volume => testvolume,
+	    status => "ready",
+	    # override defaults
+	    @_
+	);
+
+	return HTFeed::Queue->new->enqueue(%params);
+    }
+
+    sub get_vol_from_queue {
+	my $volume = shift || testvolume;
+	my $status = shift || "ready";
+
+	my $sql = join(
+	    " ",
+	    "SELECT *",
+	    "FROM   feed_queue",
+	    "WHERE  namespace = ?",
+	    "AND    id        = ?",
+	    "AND    pkg_type  = ?",
+	    "AND    status    = ?"
+	);
+
+	get_dbh()->selectrow_hashref(
+	    $sql,
+	    {},
+	    $volume->get_namespace,
+	    $volume->get_objid,
+	    $volume->get_packagetype,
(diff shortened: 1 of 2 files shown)


t/enqueue.t
```diff
@@ -0,0 +1,186 @@
...
+	enqueue_testvolume(status => 'punted');
+
+	# -u to reuse a punted item,
+	# only works in combination with a reset flag (-r/-R).
+	# (in this case i'm also re-routing stderr to devnull,
+	# because i'm doing such a bad thing that enqueue.pl
+	# is going to print its whole pod2usage to stderr, and I don't want it)
+	enqueue(
+            '-v -u -1 simple test test 2>/dev/null',
+            qr/empty result/,
```
moseshll, line 158: I'm unclear on what this actually tests. It seems to be looking for zero instances of "empty result" from STDOUT. Can you use an anchored regex to assert STDOUT is getting an empty string? Or perhaps look for the pod2usage string instead (instead of sending it to the bit bucket).

t/queue.t
```diff
@@ -1,391 +1,397 @@
...
+            };
+            it "accepts a priority" => sub {
+                my $priority = HTFeed::Queue::QUEUE_PRIORITY_LOW;
+                enqueue_testvolume(status => 'punted');
+                queue_reset(priority => $priority);
+                my $job = HTFeed::Bunnies->new()->next_job(RECV_WAIT);
+                is($job->{msg}{props}{priority}, $priority);
+            };
+
+            it "resets the priority in the database";
```
moseshll: Would like to see a comment. Should this be an `xit`? (I see this is not a new addition, still, is it a TODO?)

t/enqueue.t
```diff
@@ -0,0 +1,186 @@
...
+            $status
+        );
+    }
+
+    # Run a commandline invocation and check output against provided rx
+    sub enqueue {
+        my $cmd                = shift;
+        my $match_rx           = shift;
+        my $expect_match_count = shift;
+	my $verbose            = shift || 0;
```
moseshll, line 77: `0x09` rears its head again thanks to my BBEdit settings. To wit, whatever you did in previous PRs to get tabs expanded to spaces should be done again. I think the `my $verbose` line is the first one in the file where tab indentation starts showing up.

t/queue.t
```diff
@@ -1,391 +1,397 @@
...
+        get_dbh()->do("DELETE FROM feed_queue_disallow");
+        $testlog->reset;
+    };
 
-          ok(!HTFeed::Bunnies->new->next_job(RECV_WAIT));
+    sub fake_bibdata {
+        my $volume = shift;
+        my $sql = join(
+            " ",
+            "REPLACE INTO feed_zephir_items",
```
moseshll: I wish there were a satisfying way to stick SQL multilines in source without jumping thru hoops. I hate all of them. Have you tried using a heredoc, and comparing side by side? It's what I have tentatively settled on with Ruby. There's a trick in the Perl Cookbook for cleaning up the indentation, using `s/^\s+//gm` although for SQL I don't think it's really needed except for debugging.

moseshll: (Note, CRMS still uses a lot of
```perl
my $sql = "SELECT something FROM blah" +
  " INNER JOIN spam" +
  " ON ...";
```
and getting rid of it is my life's ambition.)

moseshll (approved):
Just fiddly comments and musings. Except for tab thing which as expected I am again being pedantic about. APPROVE in anticipation of a grand tab-squishing party.

EDIT: I forgot to mention that I haven't been able to reliably run the `feed` test suite over the past years, because I get many failures along the lines of...
```
#   Failed test 'bin/enqueue.pl can queue if there is bib data' by dying:
#     Can't locate object method "get" via package "HTFeed::Namespace::TEST"
#     at /usr/local/feed/lib/HTFeed/Namespace.pm line 33.
```

t/enqueue.t
mwarin: When faced with "how do change this function to expect zero results" I weighed between `expect 1 case of /^$/` or `expect 0 cases of "empty result"`. Neither really satisfy.

PR #150, opened 2024-08-26
bugfix for DEV-1337

mwarin:
This bug was caused by a change in return value semantics and an incomplete handling of that change.

The return value in question used to be an error code that went straight to error reporting unless zero.

Original, paraphrased:
```
system("foo") and  # 0 good, 1+ bad
report_error();
```

I changed the return value to be a success value instead, while putting the system call inside a wrapper class, but didn't fix the handling of the value in one of the classes using the new wrapper. This resulted in success going straight to error reporting.

I've now made the proper changes to the handling of that return value, so that success counts as success again.
It's a bit wordier but I prefer that.

Improved, paraphrased:
```
$success = wrapper_to_system_call("foo"); # 0 bad, 1 good
report_error() if !$success;
```

### DIFF
--- a/lib/HTFeed/Stage/ImageRemediate.pm
+++ b/lib/HTFeed/Stage/ImageRemediate.pm
@@ -818,13 +818,18 @@ sub expand_lossless_jpeg2000 {
 
                 # Single quality level with reqested PSNR of 32dB. See DEV-10
                 $start_time = $self->{job_metrics}->time;
-                HTFeed::Image::Grok::compress("$path/$tiff", "$path/$jpeg2000_remediated")
-                and $self->set_error(
-                    "OperationFailed",
-                    operation => "grk_compress",
-                    file      => "$path/$tiff",
-                    detail    => "grk_compress returned $?"
+                my $grk_compress_success = HTFeed::Image::Grok::compress(
+                    "$path/$tiff",
+                    "$path/$jpeg2000_remediated"
                 );
+                if (!$grk_compress_success) {
+                    $self->set_error(
+                        "OperationFailed",
+                        operation => "grk_compress",
+                        file      => "$path/$tiff",
+                        detail    => "grk_compress returned $?"
+                    );
+                }
                 $labels = {
                     converted => "tiff->jpeg2000",
                     tool      => 'grk_decompress'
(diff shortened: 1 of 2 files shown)


aelkiss (changes requested):
Looking at https://github.com/hathitrust/feed/blob/main/lib/HTFeed/Image/Grok.pm#L61 I think this is probably not correct, and I think 0 is actually an error (since we are already inverting the result code there.)

I wonder if it would make sense to do the error logging in `HTFeed::Image::Grok`. Then we could:
* Recording the actual exit code in the error (I think we were doing that before, but now it gets cast to a boolean and lost)
* Consider capturing the output from grok and logging that in the error (e.g. using backticks, and not redirecting the output to /dev/null)

You would likely need to do the same for the other command in `Grok.pm`, and should consider it for `Magick.pm` as well.

mwarin:
I still think this is the fix, and I hope this convinces you.

I added a test showing that `Image::Grok::compress` upon failure returns 0 and sets `$?`, and that `$?` is available to the code that calls `Image::Grok::compress`.

The real issue was that I missed one case where `ImageRemediate` would jump into set_error if the returned value was truthy. The returned value used to be a system error code (0=good, 1=bad) and is now a "success code" (0=bad, 1=good).

```perl
# Image::Grok::compress 
my $sys_ret_val = system(...); # 0 if "good"
return !$sys_ret_val;                # invert and return !0 if "good"

#####  
# ImageRemediate before PR#150
HTFeed::Image::Grok::compress("$path/$tiff", "$path/$jpeg2000_remediated") # go to next clause if "good"
and
$self->set_error(...) # goes to set error if "good" because "and"

#####

# ImageRemediate after PR#150
my $grk_compress_success = HTFeed::Image::Grok::compress(
    "$path/$tiff",
    "$path/$jpeg2000_remediated"
);
if (!$grk_compress_success) { # goes to set_error if "bad"
    $self->set_error(...)
}
```

The `0` in `grk_compress returned 0` is the error code (`$?`), meaning `Image::Grok::compress` didn't think there was an error, and we ended up setting the error because the return value of the function was 1 and the condition was set wrong to deal with that.

aelkiss (approved):
You're right. I initially thought it was fine, then second-guessed myself, and re-reading your reasoning I believe this is correct.

I do think we should capture the output from grok, but I added that as a new issue (https://github.com/hathitrust/feed/issues/151)

PR #153, opened 2024-09-20
investigating tto-261

mwarin:
See https://hathitrust.atlassian.net/browse/TTO-261

Draft, and possibly never anything more.
Just trying to figure out what is happening with these jpeg2000s that have duplicate sets of resolution tags.

Added a volume to test (under `volumes_to_test`) with 10 pages that should not be merged to main.

```
 docker compose run --rm validate
```


mwarin:
This issue is likely not solved here or this way.

Issue #156, opened 2024-11-07
strip BOM when reading checksum.md5


mwarin:
UTulsa sent us a `checksum.md5` in UTF8, with a BOM. 

Nothing says you can't do that, but we're not dealing with it properly.

PR #157, opened 2024-11-07
open checksum file in utf8 mode and strip BOM

mwarin:
Fix for issue #156, triggered by https://hathitrust.atlassian.net/issues/GS-11436.

### DIFF
--- a/lib/HTFeed/Volume.pm
+++ b/lib/HTFeed/Volume.pm
@@ -286,9 +286,13 @@ sub get_checksum_md5 {
 	    $self->set_error("MissingFile", file => $checksum_file);
 	}
 
-	open(FILE, $checksum_path) or die("Can't open $checksum_path: $!");
+        # we don't expect a BOM in checksum files, but get it sometimes anyways
+        # so it's best to open files as if they were UTF8 and strip bom if seen
+        use open ':std', ':encoding(UTF-8)';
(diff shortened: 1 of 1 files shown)


aelkiss (approved):
I don't love having to do it, but I think the fix is fine, and it's good to be generous in what we accept. It might be good to ensure that the `use open ...` is lexically scoped to this block and won't inadvertently break stuff elsewhere?

mwarin:
I think this proves the lexicalness of its scope:

```perl
use strict;                              
use warnings;                            
                                         
no_pragma();                             
use_pragma();                            
no_pragma();                             
                                         
# Should print the unicode as expected   
sub use_pragma {                         
    use open ':std', ':encoding(UTF-8)'; 
    print "Using pragma\n";              
    open(F, "utf_chars.txt") or die;     
    while (<F>) {                        
        print;                           
    }                                    
    close(F);                            
    print "-----\n";                     
}                                        
                                         
# Should print the unicode as garbled goo
sub no_pragma {                          
    print "NOT using pragma\n";          
    open(F, "utf_chars.txt") or die;     
    while (<F>) {                        
        print;                           
    }                                    
    close(F);                            
    print "-----\n";                     
}                                        
```

Just make a utf_chars.txt and put some `€`s in it or something, and watch it print the same stuff differently depending on the scope.

PR #160, opened 2025-01-15
Update Dockerfile for progress_tracker v0.11.0

aelkiss:
Use the latest progress tracker package.

Not sure if we really need to be pinning here or if it's more trouble than it's worth.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -79,7 +79,7 @@ COPY ./Makefile.PL $FEED_HOME/Makefile.PL
 
 RUN cpanm --notest -l /extlib \
   https://github.com/hathitrust/metslib.git@v1.0.1 \
-  https://github.com/hathitrust/progress_tracker.git@v0.9.0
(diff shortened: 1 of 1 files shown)


aelkiss:
@mwarin If/when you have a chance could you update feed_internal with this as well (i.e. release a new version of that referencing the updated version of feed?)

mwarin (approved):
Looks good. Please merge at will, and I'll take it from there (i.e make a new feed release, and a new feed_internal release referencing the new feed release)

aelkiss:
Unfortunately we need to do another update and release again, because progress tracker 0.11.0 didn't do what we needed -- but it isn't terribly urgent for ingest.

mwarin:
> Unfortunately we need to do another update and release again, because progress tracker 0.11.0 didn't do what we needed -- but it isn't terribly urgent for ingest.

Do we want https://github.com/hathitrust/progress_tracker/releases/tag/v0.11.1, then?

PR #163, opened 2025-06-10
ensuring resolution does not contain `ppi`

mwarin:
Fix for https://hathitrust.atlassian.net/browse/GS-13230

... in which some of `bc`'s IA items were failing because they fall back on an invalid resolution value from the `$volume`.
Single line fix that removes `ppi` from resolution values.

This should fail on the main branch:
```
$ git checkout main
$ docker compose run --rm test perl bin/ingest_ia_dev.pl realceduladesuma00spai_0 test ark:/13960/s25zng884bb
[...]
Error setting new tag Resolution => 400ppi: Not a floating point number for MIE-Image:Resolution
```
... but should succeed on this PR branch:

```
$ git checkout GS-13230-ppi-resolution-fix
$ docker compose run --rm test perl bin/ingest_ia_dev.pl realceduladesuma00spai_0 test ark:/13960/s25zng884bb
```

### DIFF
--- a/lib/HTFeed/PackageType/IA/ImageRemediate.pm
+++ b/lib/HTFeed/PackageType/IA/ImageRemediate.pm
@@ -29,6 +29,7 @@ sub run {
     if (not defined $resolution or !$resolution) {
 	$resolution = $volume->get_meta_xpc()->findvalue("//ppi");
     }
+    $resolution =~ s/ppi//;
 
     # decompress any lossless JPEG2000 images
     my @jp2 = glob("$preingest_dir/*.jp2");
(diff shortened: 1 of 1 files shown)


mwarin:
@aelkiss I would like a chance to run this through with you once or twice. I think there are dueling red herrings:

* can't run it as namespace:`test` because it's going to look for things in zephir and fail
* can't run it as namespace:`bc` in dev because I don't know how to make real namespaces work in dev
* once i manage to get it in zephir, the arkid on https://zephir.cdlib.org/api/item/bc.ark:/13960/s25zng884bb does not match the arkid realceduladesuma00spai_0

aelkiss (approved):
After testing with the correct IA ID matching this arkid, everything looks good. We added ETT-340 as a potential future step to make testing this kind of thing easier.

PR #165, opened 2025-06-30
ETT-374

mwarin:
* adding dir for storing extracted bibdata


aelkiss:
This should be fine as-is. I will go ahead and merge/release/etc so that we can get tests for https://github.com/hathitrust/feed_internal/pull/108 passing in github.

PR #167, opened 2025-09-18
ETT-571: handle empty DateTime in RGB TIFFs 

aelkiss:
This ended up being a bit tricky to track down because the error didn't give a useful stack trace, and the culprit is buried in the rather long `convert_tiff_to_jpeg2000` method.

This was indeed a one-line fix (https://github.com/hathitrust/feed/pull/167/commits/6b1f1167e10fb0a4c820f272500723d84f973774#diff-0d491008f8069f2e647a11563976e76d9a8fd71e522c3efbb3b9384b58b93ae6R175), but getting there was a bit tricky in terms of figuring out exactly what triggered the issue, creating a test fixture, and then actually finding the source of the problem.

It may be worth thinking a bit about what we do with ImageRemediate in the future – in particular if we want to make major changes to this in the future we should do some refactoring and unit testing there first.

https://github.com/hathitrust/feed/commit/6b1f1167e10fb0a4c820f272500723d84f973774 is the commit with the actual changes. Some comments inline.


t/local_ingest.t
```diff
@@ -170,6 +170,24 @@ describe "HTFeed::PackageType::Simple" => sub {
       $validate->run();
       ok($validate->succeeded());
     };
+
+    it "remediates empty DateTime in a bitonal TIFF file" => sub {
```
aelkiss, line 174: This test (with its associated fixture) passed without any changes to the `ImageRemediate` code, but I think it's worth keeping as a regression test.

lib/HTFeed/Stage/ImageRemediate.pm
```diff
@@ -171,8 +171,9 @@ sub copy_old_to_new($$$) {
 
   my $oldValue = $self->{oldFields}->{$oldFieldName};
   if (
-    defined $self->{oldFields}->{$oldFieldName} and
-    not defined $self->{newFields}->{$newFieldName}
+    defined $self->{oldFields}->{$oldFieldName} 
+      and $self->{oldFields}->{$oldFieldName} ne ''
```
aelkiss, line 175: This method (`copy_old_to_new`) is called here: https://github.com/hathitrust/feed/blob/6b1f1167e10fb0a4c820f272500723d84f973774/lib/HTFeed/Stage/ImageRemediate.pm#L1165 but ExifTool didn't like being given an empty value for the new field value. This guards against that particular possibility. It might help avoid certain other kinds of oddities related to `copy_old_to_new` but I don't know if we've run into that exactly.

liseli (approved):
These changes look good. I appreciate seeing the different tests that consider various scenarios for image validation.

PR #168, opened 2025-09-26
ETT-574: Deposit to new storage with versitygw S3 gateway

aelkiss:
This implements a "PairtreeObjectStore" which we can use with [versitygw](https://github.com/versity/versitygw/) - a filesystem to S3 gateway - to deposit to both the local & remote storage without needing to do NFS mounts. PageTurner, etc, will still access the filesystem with NFS, and it will be laid out in more or less the same way, so we do want to use the "pairtree path".

Not yet implemented is anything recording this deposit; in the next PR, I'd like to add a column to feed_audit and separately record each deposit that ingest does.

In addition, this removes minio (another thing that implements S3) that we were using to mock depositing to glacier deep archive --  we don't need two separate test S3 implementations here.

This also tests that symlinks will work correctly with versitygw. For existing stuff the terminal directory under sdr1 is a symlink to one of sdr2 .. sdr24; this tesst that we can write through symlinks with versitygw -- e.g. that we can write using that symlink in the path and update the existing zip/mets. For new material, we'll just deposit right under sdr1, and not bother with the symlinks. Over time, we can (separately from this process) work towards removing the existing symlinks.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,9 +25,9 @@ services:
     command: prove
     depends_on:
       mariadb: *healthy
-      minio: *healthy
       pushgateway: *healthy
       rabbitmq: *healthy
+      versitygw: *healthy
 
   # Note: for permissions purposes this does NOT bind in the local development
   # environment, so local changes after running docker compose build will NOT
@@ -36,6 +36,7 @@ services:
     build: .
     volumes:
       - ./clamav:/var/lib/clamav
+      - ./var/vgw:/usr/local/feed/var/vgw
     environment:
       - HTFEED_CONFIG=/usr/local/feed/etc/config_test.yml
       - FEED_HOME=/usr/local/feed
@@ -53,9 +54,9 @@ services:
     command: cover -test -report Coveralls -make 'prove; exit $?'
     depends_on:
       mariadb: *healthy
-      minio: *healthy
       pushgateway: *healthy
       rabbitmq: *healthy
+      versitygw: *healthy
 
   ingest:
     build: .
@@ -119,28 +120,20 @@ services:
       <<: *healthcheck-defaults
       test: ["CMD", "healthcheck.sh", "--su-mysql", "--connect", "--innodb_initialized"]
 
-  # S3 compatible object storage
-  minio:
-    image: minio/minio
-    restart: always
-    environment:
-      MINIO_ACCESS_KEY: TESTINGACCESSKEY
-      MINIO_SECRET_KEY: testingsecretkey
-    command: server /data
-    healthcheck:
-      <<: *healthcheck-defaults
-      test: timeout 5s mc ready local
-
   # S3 -> filesystem gateway
   versitygw:
+    user: "1000:1000"
     image: versity/versitygw
     restart: always
     environment:
       ROOT_ACCESS_KEY: TESTINGACCESSKEY
       ROOT_SECRET_KEY: testingsecretkey
     volumes:
-      - ./var/vgw:/vgw
-    command: posix /vgw
+      - ./var/vgw:/usr/local/feed/var/vgw
+    command: --health /health posix /usr/local/feed/var/vgw
+    healthcheck:
+      <<: *healthcheck-defaults
+      test: [ "CMD", "wget", "--quiet", "--tries=1", "-O", "/dev/null", "http://127.0.0.1:7070/health" ]
 
   pushgateway:
     image: prom/pushgateway
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -131,6 +131,17 @@ services:
       <<: *healthcheck-defaults
       test: timeout 5s mc ready local
 
+  # S3 -> filesystem gateway
+  versitygw:
+    image: versity/versitygw
+    restart: always
+    environment:
+      ROOT_ACCESS_KEY: TESTINGACCESSKEY
+      ROOT_SECRET_KEY: testingsecretkey
+    volumes:
+      - ./var/vgw:/vgw
+    command: posix /vgw
+
   pushgateway:
     image: prom/pushgateway
     command:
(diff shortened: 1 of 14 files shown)


aelkiss:
@rrotter Not sure how much review you might want to do on this, but figured I'd point it out at least.

aelkiss:
Tests are failing; it looks like versitygw isn't starting. I'll take a look at that in the docker compose.

aelkiss:
Tests are still failing I suspect due to differing permission schemes under github actions as locally. I will see what makes sense to do.

aelkiss:
Tests are working in GHA now.

I've also added logic to record deposits to a new `feed_storage` table; this is separate from the existing `feed_audit` table which I think it doesn't make sense to reuse because of the extra stuff recorded there.

One thing that comes to mind is that we may want to extract a separate subclass of `ObjectStore` with the logic used for backup so we can have only the stuff that's common in the superclass.

docker-compose.yml
```diff
@@ -119,17 +120,20 @@ services:
...
     restart: always
     environment:
-      MINIO_ACCESS_KEY: TESTINGACCESSKEY
-      MINIO_SECRET_KEY: testingsecretkey
-    command: server /data
+      ROOT_ACCESS_KEY: TESTINGACCESSKEY
+      ROOT_SECRET_KEY: testingsecretkey
+    volumes:
+      - ./var/vgw:/usr/local/feed/var/vgw
+    command: --health /health posix /usr/local/feed/var/vgw
```
moseshll: I had issues running this locally: vgw was unhealthy complaining "2025/10/02 17:33:13 xattr check failed: xattrs are not supported on this filesystem". Saw this with the Otis branch too. Passing `--nometa` fixed this for Otis. I don't know if this implementation is depending on xattrs for the tests to pass, Otis by contrast is using vgw read-only and it doesn't seem to matter.

I did an A/B test on `main` vs this branch, since I always get failing tests -- on this branch I got
```
t/storage_object_store.t ........... 7/? 
#   Failed test 'metadata for md5 checksum'
#   at t/storage_object_store.t line 75.
#          got: undef
#     expected: 'LUDWXBrs2Fez94DoW8m9kg=='
```
which suggest we need that xattr stuff.

I don't know if it's a priority to get this running on non-ext3 (or whatever) filesystems since the feed test suite has always been ornery IMHO.

aelkiss: I can try on Mac and with `--no-meta`.

aelkiss: I suspect the xattrs may be needed to store checksums though.

moseshll (changes requested):
See comment on docker-compose about my filesystem woes. I don't see anything wrong with the code changes but it would be good to address this issue.

docker-compose.yml
moseshll: Yeah I edited my comment after comparing test outputs.

moseshll: I tested with a sidecar approach (https://github.com/versity/versitygw/wiki/POSIX-metadata) and didn't get the above error but a test in t/storage_object_store.t still failed (HTFeed::Storage::ObjectStore with encryption enabled stores the mets and encrypted zip), so I dunno.

This was my change:
```
versitygw:
    user: "1000:1000"
    image: versity/versitygw
    restart: always
    environment:
      ROOT_ACCESS_KEY: TESTINGACCESSKEY
      ROOT_SECRET_KEY: testingsecretkey
    volumes:
      - ./var/vgw:/usr/local/feed/var/vgw
      - ./var/sidecar:/usr/local/feed/var/sidecar
    command: --health /health posix --sidecar /usr/local/feed/var/sidecar /usr/local/feed/var/vgw
    healthcheck:
      <<: *healthcheck-defaults
      test: [ "CMD", "wget", "--quiet", "--tries=1", "-O", "/dev/null", "http://127.0.0.1:7070/health" ]
```
(had to create `./var/sidecar`)

aelkiss: > I don't know if it's a priority to get this running on non-ext3 (or whatever) filesystems since the feed test suite has always been ornery IMHO.

I'd love to have it all working consistently, especially since now I'm now the only developer in ETT primarily running tests under Linux. I do have a mini I can try some of this on and at least try to triage.

rrotter: fwiw, I successfully tested versitygw on my Mac running in /tmp (more or less using their quickstart docs verbatim). Did not try with Docker, just installed from brew.

aelkiss: Re: the failing tests I am seeing the same thing. I see a lot of errors like:

"gpg: error running '/usr/bin/gpg-agent': exit status 2"

I suspect this is related to the test failures.

I also note that using the sidecar leaves a lot of junk in the sidecar directory lying around after test. I will see if I can at least use a directory or volume that doesn't leave stuff in the repository filesystem, since we aren't separately checking any of that from other containers.

aelkiss: I'm running into permissions & other weird issues if I try to have the sidecar somewhere else or in a regular Docker volume because I'm not running versitygw as root, so I think we'll have to live with accumulating junk in the temp directory. I can at least make sure it's gitignored.

aelkiss: I think the gpg issue also has to do with the Mac filesystem and/or mapping of Unix sockets. In any case I can move the gpg directory to /tmp/gnupg (as we do in production) and that's fine.

aelkiss:
@moseshll All the tests on this branch pass for me via Docker on Mac now. Could you give it another try?

aelkiss:
Well now tests are failing in GitHub.. but I bet I know why & it has to do with permissions for the metadata sidecar...

aelkiss:
OK tests are passing in github, locally for me in Docker running in Linux, and locally for me in Docker running in Mac OS. @moseshll if they work for you in Mac OS I think we can merge this!

moseshll (approved):
Tests pass and output generally matches what we're getting from the CI tests here. Hearty APPROVE

PR #169, opened 2025-10-06
Fix failing tests on Mac

aelkiss:
The Mac filesystem is case-preserving but not case sensitive, and t/lib/HTFeed/Namespace/Test.pm appears to conflict with lib/HTFeed/Namespace/TEST.pm. This appears to fix that issue.

@moseshll I'm opening a separate draft PR for this so you can test it independently of the work on #168, but I am going to pull this commit in over there so I can test the versitygw issues without other things interfering.


aelkiss:
I'm a little surprised we aren't running into the gpg issue on Mac here as well, but I think I have a fix.

aelkiss:
Included in #168

PR #170, opened 2025-10-30
ETT-780: Bucket root contains obj dir

aelkiss:
@rrotter When testing before I had assumed that the root of the bucket corresponded to the 'obj' dir. However in the setup you have the bucket points to `/sdr1` rather than `/sdr1/obj`. That works but needed some adjustment in the ingest code. Let me know if you think it would be preferable to have the bucket point to `/sdr1/obj` -- I don't have a strong preference.

### DIFF
--- a/lib/HTFeed/Storage/PairtreeObjectStore.pm
+++ b/lib/HTFeed/Storage/PairtreeObjectStore.pm
@@ -12,7 +12,7 @@ sub object_path {
   my $self = shift;
 
   return sprintf(
-      '%s/%s%s/',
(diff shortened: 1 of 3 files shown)


rrotter:
I chose that because I'd thought you'd want paths to be similar the existing /sdr1 mounts on apache hosts. I actually prefer linking a lever deeper to `/sdr1/obj`. See: https://github.com/hathitrust/truenas-stacks/pull/1

aelkiss:
Superceded by https://github.com/hathitrust/truenas-stacks/pull/1

PR #175, opened 2026-03-03
ETT-1288 Run fixity check on new storage

moseshll:
- Add `storage_name` column to `feed_audit_detail`
- Add `truenas_audit.pl` script based on `main_repo_audit.pl`
- `HTFeed/StorageAudit.pm` updated to provide `feed_audit_detail.storage_name`


t/truenas_audit.t
```diff
@@ -0,0 +1,175 @@
...
+  
+  it "records a spurious file but ignores pre-uplift METS" => sub {
+    my $storage_name = 's3-truenas-macc';
+    my $objid = 'test';
+    # Add a silly file and a pre-uplift file
+    my $pt_objid = s2ppchars($objid);
+    my $pt_path = id2ppath($objid);
+    foreach my $ext (('silly', 'pre_uplift.mets.xml')) {
+      my $path = "/tmp/sdr1/obj/test/$pt_path$pt_objid/" . "$objid.$ext";
+      open(my $fh, '>', $path) or die "open file $path failed: $!";
```
moseshll: FIXME: should suffice to just `touch` the files, we don't care about contents.

bin/audit/truenas_audit.pl
```diff
@@ -0,0 +1,298 @@
...
+use HTFeed::METS;
+use HTFeed::Namespace;
+use HTFeed::PackageType;
+use HTFeed::RepositoryIterator;
+use HTFeed::Volume;
+use HTFeed::VolumeValidator;
+
+
+# FIXME: is this needed?
+my $tombstone_check = "select is_tombstoned from feed_audit where namespace = ? and id = ?";
```
aelkiss, line 26: I'm not certain. We haven't tombstoned anything in many years. Worth checking what has `is_tombstoned=1`, though.

bin/audit/truenas_audit.pl
```diff
@@ -0,0 +1,298 @@
...
+      $mets_seconds = ( stat($metsfile) )[9];
+      $metssize     = -s $metsfile;
+      $metsdate     = strftime( "%Y-%m-%d %H:%M:%S",
+        localtime( ( stat($metsfile) )[9] ) );
+    }
+
+    my $last_touched = $zip_seconds;
+    $last_touched = $mets_seconds if defined $mets_seconds and (not defined $zip_seconds or $mets_seconds > $zip_seconds);
+
+    # FIXME: I don't know if this is needed and if it is this is old code from main_repo_audit.pl so it needs fixin'
```
aelkiss: I think this is worth keeping for now to verify symlinks from sdr1 to the "real" storage location are intact, although we can remove the `$last_touched` check.

lib/HTFeed/Storage/LocalPairtree.pm
```diff
@@ -138,19 +138,24 @@ sub record_audit {
...
-    "insert into feed_audit (namespace, id, sdr_partition, zip_size, zip_date, mets_size, mets_date, lastchecked, lastmd5check, md5check_ok) \
-    values(?,?,?,?,?,?,?,CURRENT_TIMESTAMP,CURRENT_TIMESTAMP,1) \
+    "insert into feed_audit (namespace, id, storage_name, sdr_partition, zip_size, zip_date, mets_size, mets_date, lastchecked, lastmd5check, md5check_ok) \
+    values(?,?,?,?,?,?,?,?,CURRENT_TIMESTAMP,CURRENT_TIMESTAMP,1) \
     ON DUPLICATE KEY UPDATE sdr_partition = ?, zip_size=?, zip_date =?,mets_size=?,mets_date=?,lastchecked = CURRENT_TIMESTAMP,lastmd5check = CURRENT_TIMESTAMP, md5check_ok = 1";
 
     # TODO populate image_size, page_count
 
+    # FIXME: is this right?? should we force it to one of {s3-truenas-ictc, s3-truenas-macc}?
+    my $storage_name = $self->{storage_name} || 'LocalPairtree';
```
aelkiss: Probably? Or rather, we can bail out if `storage_name` isn't set?

t/repository_iterator.t
```diff
@@ -0,0 +1,88 @@
...
+  spec_helper 'storage_helper.pl';
+  local our ($tmpdirs, $testlog);
+  
+  sub make_sdr_entry {
+    my $namespace = shift;
+    my $objid = shift;
+
+    my $pt_objid = s2ppchars($objid);
+    my $pt_path = id2ppath($objid);
+    my $full_path = "/tmp/sdr1/obj/$namespace/$pt_path" . $pt_objid;
```
aelkiss, line 21: may want to have the root under something using `tempdir` from `File::Temp`, rather than right under `/tmp`, so we don't have leftover stuff from the previous run

t/truenas_audit.t
```diff
@@ -0,0 +1,173 @@
...
+    my $namespace = 'test';
+    my $objid = 'test';
+    my $storage = local_storage($namespace, $objid);
+    $storage->stage;
+    $storage->make_object_path;
+    $storage->move;
+    my $pt_objid = s2ppchars($objid);
+    my $pt_path = id2ppath($objid);
+    # truenas_audit.pl can infer its sdr partition when it isn't at the root of the
+    # filesystem but the `tmpdirs` random names will throw it off completely. Hence we
```
aelkiss: Can we have e.g. `/tmp/SOMERANDOMSTUFF/sdr1`?

aelkiss (commented):
I've looked at this some; I want to look at the tests more, not sure I'll get to that today though.

t/truenas_audit.t
```diff
@@ -0,0 +1,173 @@
...
+  after each => sub {
+    File::Path::remove_tree('/tmp/sdr1');
+    File::Path::remove_tree('/tmp/obj_link');
+    get_dbh->prepare('DELETE FROM feed_storage')->execute;
+    get_dbh->prepare('DELETE FROM feed_audit_detail')->execute;
+  };
+
+  foreach my $storage_name (('s3-truenas-macc', 's3-truenas-ictc')) {
+    it "writes to feed_storage" => sub {
+      `bin/audit/truenas_audit.pl --md5 --storage_name $storage_name /tmp/sdr1`;
```
aelkiss: Would it make sense to wrap the contents of `truenas_audit.pl` and do a `main unless caller` kind of thing at the end, to make it easier to test?

t/truenas_audit.t
```diff
@@ -0,0 +1,173 @@
...
+    $storage->stage;
+    $storage->make_object_path;
+    $storage->move;
+    my $pt_objid = s2ppchars($objid);
+    my $pt_path = id2ppath($objid);
+    # truenas_audit.pl can infer its sdr partition when it isn't at the root of the
+    # filesystem but the `tmpdirs` random names will throw it off completely. Hence we
+    # copy to a location where we can put "sdr1" in the path.
+    File::Path::make_path('/tmp/sdr1/obj');
+    `cp -r $tmpdirs->{obj_dir}/* /tmp/sdr1/obj/`;
```
aelkiss: Might be worth using `system` rather than backticks since we're throwing away the return value; that also has the benefit of allowing stdout to go to console which could help in debugging issues with tests.

aelkiss (commented):
Tests look good; see comments about the backticks. Not critical, but maybe worth considering.

bin/audit/truenas_audit.pl
moseshll: There are 13 tombstoned volumes in feed_audit, none with a `mets_date` more recent than Feb 2014. That does answer my question, though, and I will leave the query in place.

t/truenas_audit.t
moseshll: Yes, fortunately (see previous comment).

t/repository_iterator.t
moseshll: I've done this in the most recent commit. Alas, the test config enforces `/tmp/obj_link` as the link directory, but `/tmp/sdr1` is no longer used.

bin/audit/truenas_audit.pl
moseshll: Code fixed up and enabled, tests added.

t/truenas_audit.t
moseshll: Backticks have been changed to `system`

lib/HTFeed/Storage/LocalPairtree.pm
moseshll: I've added a bug fix (it's self->name not self->storage_name) and it just bails out without error if the LocalPairtree name isn't set (which I think it always will be, from the way the config works).

That said, I will have some comments below about whether this change is needed.

t/truenas_audit.t
moseshll: I'm far enough along at this point that the test lashup is cooperating. If this were not an alleged one-off I would seriously considering doing so.

moseshll:
@aelkiss Since the scope of this has changed to focus on `feed_storage` instead of `feed_audit` I would suggest backing out changes to the `feed_audit` DB schema on this branch. I suspect the changes may still be disruptive for feed_internal but if this is to be the MVP for this round then I'd like to return to `feed_audit` after getting some utility from what's here.
Thoughts?

bin/audit/truenas_audit.pl
```diff
@@ -0,0 +1,281 @@
...
+    my $metssize;
+
+    if ( -e $metsfile ) {
+      $mets_seconds = ( stat($metsfile) )[9];
+      $metssize     = -s $metsfile;
+      $metsdate     = strftime( "%Y-%m-%d %H:%M:%S",
+        localtime( ( stat($metsfile) )[9] ) );
+    }
+
+    my $last_touched = $zip_seconds;
```
aelkiss: `$last_touched` doesn't appear to be used

bin/audit/truenas_audit.pl
```diff
@@ -0,0 +1,281 @@
...
+      if ( defined $link_target and $link_target ne $path ) {
+        set_status( $namespace, $objid, $storage_name, $path, "SYMLINK_INVALID",
+          $link_target || '<undef>' );
+      }
+
+    }
+
+
+    #insert
+    execute_stmt(
```
aelkiss: Maybe wait to do this until after the file checks? Maybe worth comparing to the backup audits to see behavior -- that is, do we update lastchecked when we see it or after we verify some part of it?

t/truenas_audit.t
```diff
@@ -0,0 +1,269 @@
...
+      my $temp_sdr_path = temp_sdr_path;
+      system("bin/audit/truenas_audit.pl --md5 --storage_name $storage_name $temp_sdr_path");
+      my $db_data = get_feed_storage_data('test', 'test', $storage_name);
+      is(scalar(@$db_data), 1, 'with only one initial entry');      
+      is($db_data->[0]->{namespace}, 'test', 'correct namespace');
+      is($db_data->[0]->{id}, 'test', 'correct id');
+      is($db_data->[0]->{storage_name}, $storage_name, 'correct storage_name');
+      ok($db_data->[0]->{zip_size} > 0, 'nonzero zip_size');
+      ok($db_data->[0]->{mets_size} > 0, 'nonzero mets_size');
+      ok(!defined $db_data->[0]->{saved_md5sum}, 'not defined saved_md5sum');
```
aelkiss: Do we want it to save this? Is it worth using if it's there? Probably fine for now to leave as-is, but maybe worth thinking about in the future or otherwise making it more consistent with what we do for the backup auditing.

etc/ingest.sql
```diff
@@ -3,6 +3,7 @@ USE `ht`;
 CREATE TABLE IF NOT EXISTS `feed_audit` (
   `namespace` varchar(10) NOT NULL,
   `id` varchar(30) NOT NULL,
+  `storage_name` varchar(32) NOT NULL,
```
aelkiss: I think it makes sense to go ahead and roll back these changes

lib/HTFeed/Storage/LocalPairtree.pm
```diff
@@ -138,19 +138,21 @@ sub record_audit {
     my ($sdr_partition) = ($path =~ qr#/?sdr(\d+)/?#);
 
     my $stmt =
-    "insert into feed_audit (namespace, id, sdr_partition, zip_size, zip_date, mets_size, mets_date, lastchecked, lastmd5check, md5check_ok) \
-    values(?,?,?,?,?,?,?,CURRENT_TIMESTAMP,CURRENT_TIMESTAMP,1) \
+    "insert into feed_audit (namespace, id, storage_name, sdr_partition, zip_size, zip_date, mets_size, mets_date, lastchecked, lastmd5check, md5check_ok) \
```
aelkiss: Likewise here.

However, we also need to think about once we turn off the old storage that uses `LocalPairtree`, we probably still want feed_audit populated? Maybe we need to mark one storage as "primary" for the purposes of `feed_audit`, or have some kind of view. Regardless, we don't need to do it right now in this PR.

aelkiss (approved):
There are definitely some things here to think about in the future:

* how we use the checksums saved in `feed_storage`
* questions about `feed_audit` going forward
* in the future, merging this with `StorageAudit` and making it more testable

I think this will get us what we need for now, and we can make some follow-up issues for the remaining issues.

bin/audit/truenas_audit.pl
moseshll: It was used in some defunct recency calculations alongside the sdr1 check. Now given the heave-ho as it should have.

bin/audit/truenas_audit.pl
moseshll: I moved this a little closer to the zipcheck call. At that point we have at least checked the file count and looked for BAD_FILE, BARCODE_MISMATCH, and symlink issues.

t/truenas_audit.t
moseshll: After looking at some of the other code I changed the behavior to record the actual zip md5, so now that particular test requires the value to be defined.

etc/ingest.sql
moseshll: Changes reverted.

PR #178, opened 2026-03-27
ETT-1288 fixes for running fixity check on new storage

moseshll:
- Lessons learned from running a neutered version of truenas_audit.pl against several sdr partitions and namespaces.
  - Fix `ku01` errors resulting from including the final directory in `ppath2id`
  - Ignore `pairtree_version*` in paths considered by repo iterator
  - Fix at least two types of pairtree errors encountered with escaped barcodes
  - Allow `sdr/X/...` partitions in addition to `sdrX/...`
  - Do not follow symlinks if partition is `sdr1`
  - Tests use `sdr2` instead of `sdr1` since following symlinks on `sdr1` is verboten
  - Fix `RepositoryIterator` tests exposed as brittle under GH actions.


lib/HTFeed/RepositoryIterator.pm
```diff
@@ -113,7 +114,9 @@ sub _find_pipe {
 
   if (!$self->{find_pipe}) {
     my $find_pipe;
-    my $find_cmd = "find $self->{path} -follow -type f|";
+    # Do not follow symlinks under sdr1, okay to follow otherwise
```
aelkiss: There `SHOULD NOT` be symlinks under `/sdrN` (but `/sdrN` itself will be a symlink)

aelkiss (commented):
I'm not clear on the escaping issue, but it's probably fine. Do you have a specific example that was causing issues?

I don't think we need to follow symlinks under any conditions, but LMK if you find a counterexample.

lib/HTFeed/RepositoryIterator.pm
moseshll: That's why I added support for /sdr/1 because it was the quick fix -- otherwise I just could get `find` to come up with anything! However, I could probably normalize the `base` script parameter (`my $base = shift @ARGV`) to truenas_audit and resolve symlinks once (with `Cwd::abs_path` or such), then never pass `-follow` in the iterator.

moseshll: I'm hoping the new commit will make things more intuitive. `Cwd::abs_path` seems to do the trick over on dev-2.

bin/audit/truenas_audit.pl
```diff
@@ -68,7 +68,8 @@
   my $namespace = $obj->{namespace};
   my $objid = $obj->{objid};
   eval {
-    if ($obj->{directory_objid} ne $objid) {
+    # (Already escaped) directory name must match escaped objid
+    if ($obj->{directory_objid} ne s2ppchars($objid)) {
```
moseshll, line 73: This is the main location where the issue with escaped paths came to be exposed. I was comparing escaped objid vs unescaped objid. Unfortunately, the tests do not expose this.

aelkiss:
I've built this; going to deploy in ingest-testing and test a bit as we did before; if that looks good, I'll try running for a little bit in regular (non-"noop") mode but against the development database. If that looks good, then I'll
* merge, build, and deploy
* update database schema in production (if needed)
* try running a little bit against one /sdrN partition and also /sdr1 at each site
* if that looks good, start the full audit

aelkiss:
If I start this with an argument of either `/sdr5` or `/sdr/5`, I get tons of errors like:

```
pst 000071533759 s3-truenas-macc /sdr/5/obj/pst/pairtree_root/00/00/71/53/37/59/000071533759 CANT_LSTAT /sdr/5/obj/pst/pairtree_root/00/00/71/53/37/59/000071533759 Invalid argument
```

aelkiss:
Conversely, if I use as an argument `/sdr1/obj/coo1/pairtree_root/ar/k+/\=1/39/60/\=s/2p` (this path has some recent ingests that are not symlinks elsewhere), audit doesn't seem to find them.

I'll take a closer look and see if I can tell what needs to happen, using the setup I have in ingest-testing and just copying in an updated version of the audit script.

bin/audit/truenas_audit.pl
aelkiss: This no longer works in that it the underlying paths are like `/sdr/5`, but we need to check the symlink in `/sdr1` which then points to `/sdr5/`.

My suggestion:

```perl
$link_path =~ s/(sdr\/)?$sdr_partition/${1}1/;

...

if ( defined $link_target and Cwd::abs_path($link_target) ne $path ) {
  set_status(... , "SYMLINK_INVALID")
}
```

That is, make finding the `/sdr1` version agnostic to whether we're using `/sdr/1` or `/sdr1`, and pass `$link_target` through `Cwd::abs_path`.

I tried this in ingest testing and it seems to work OK. I'll push a commit that addresses both the issues.

aelkiss:
@moseshll See additional commit. I'll try to adjust the "detects bad symlinks" test and also add a test for the case of objids with characters that end up pairtre-escaped

PR #179, opened 2026-04-29
ETT-1473: checkpoint needs to account for storage

aelkiss:
otherwise it will only check it if it hasn't already been checked on at least one storage, but we want to check *all* the storages


aelkiss:
FYI @moseshll. I'm just going to merge this (and then do the feed release & feed_internal release...)

PR #180, opened 2026-05-01
ETT-1393 - small fixes for IA ingest

aelkiss:
* ImageRemediate: handle filenames with embedded '.' characters (previously confused file type detection for items with IA IDs containing '.' characters)

* Log grok command line at TRACE level

* Accept 1 <= DecompositionLevels <= 32 for IA material in general

### DIFF
--- a/lib/HTFeed/Image/Grok.pm
+++ b/lib/HTFeed/Image/Grok.pm
@@ -5,6 +5,7 @@ use warnings;
 
 use HTFeed::Config qw(get_config);
 use HTFeed::Image::Shared;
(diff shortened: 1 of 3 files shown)


aelkiss:
FYI @moseshll. Small enough changes I don't think it needs much of a review.  We already accept 1 - 32 for locally-digitized material (https://github.com/hathitrust/feed/blob/main/lib/HTFeed/PackageType/Simple.pm#L140), so I don't think there's any reason to insist on it for material coming through IA.

moseshll (approved):
Passes inspection. APPROVE

PR #182, opened 2026-05-05
ETT-1502: collate uses volume directly to determine repeat

aelkiss:
Currently in production, we are no longer using LinkedPairtree or LocalPairtree to deposit material. They were responsible for setting is_repeat. Now, collate checks directly.

This points to a sort of mismatch for how for writing to the repository we have abstracted the repository storage, but not for reading what is in the repository. That might be future work to separate out from Volume.

### DIFF
--- a/lib/HTFeed/Stage/Collate.pm
+++ b/lib/HTFeed/Stage/Collate.pm
@@ -26,102 +26,103 @@ HTFeed::Stage::Collate
 =cut
 
 sub storages {
-    my $self = shift;
-    return HTFeed::Storage::for_volume($self->{volume});
+  my $self = shift;
+  return HTFeed::Storage::for_volume($self->{volume});
 }
 
 sub run{
-    my $self = shift;
+  my $self = shift;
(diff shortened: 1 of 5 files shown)


aelkiss:
This removes the tests and functionality for `is_repeat` from `LinkedPairtree` and `LocalPairtree`. We might want to remove those classes entirely, although `LocalPairtree` could potentially be a base for future functionality around reading from the repository.

There is a test in `Collate` for the `is_repeat` functionality which remains the same and should cover this case:
https://github.com/hathitrust/feed/blob/795e9f105f93954e06bf1f673a6573cdc8414c17/t/collate.t#L163-L173

aelkiss:
I recommend viewing w/o whitespace changes: https://github.com/hathitrust/feed/pull/182/changes?w=1
since code formatting did a number on it..

moseshll (approved):
Wow, ignoring whitespace definitely helps. I see no issues here, APPROVE

PR #183, opened 2026-05-05
Set default build platform for build on workflow_run

aelkiss:


aelkiss:
FYI @kron-spar . This behavior confused me but then I realized we changed the default in the build action and weren't passing through the default value from here when it runs automatically (which it doesn't in most of the other repos)

PR #184, opened 2026-05-05
gha updates

aelkiss:
* tag-release shouldn't use unstable
* build should not push latest by default


aelkiss:
FYI @kron-spar. This action has drifted out of sync with other repos and I'm trying to get it more like that (while keeping the build by default on merge). In general this stack is pretty annoying to build & deploy though with the feed vs. feed_internal thing.

PR #185, opened 2026-06-04
ETT-1459: Record first ingest date

aelkiss:
This change moves the functionality for recording items in `feed_audit` to the end of the Collate stage rather than a particular storage.

It also takes the opportunity to:
* clean up some confusing configuration regarding `link_dir` and `obj_dir`
* improve testing functionality around temporary directories
* remove LinkedPairtree entirely (since we no longer deposit using symlinked anything)
* log how long each collate operation takes (ETT-824); add additional fields to info & warn level logging in collate

See comments in more detail on each commit.

I had looked into options for rolling back failed deposits to S3, but the ideas I had didn't work out (see ETT-1483).


etc/ingest.sql
```diff
@@ -5,6 +5,7 @@ CREATE TABLE IF NOT EXISTS `feed_audit` (
   `id` varchar(30) NOT NULL,
   `sdr_partition` tinyint(4) DEFAULT NULL,
   `zip_size` bigint(20) DEFAULT NULL,
+  `first_ingest_date` datetime NULL DEFAULT CURRENT_TIMESTAMP,
```
aelkiss, line 8: This is the change for recording first ingest date; the application side doesn't need to handle it directly at all beyond making sure that something is recorded in `feed_audit`

lib/HTFeed/Stage.pm
```diff
@@ -80,6 +80,8 @@ sub failed {
     return 1;
 }
 
+
+# FIXME deduplicate
```
aelkiss: Still should probably do that...

lib/HTFeed/Storage.pm
```diff
@@ -92,7 +92,7 @@ sub zip_source {
 sub encrypted_zip_staging {
     my $self = shift;
 
-    return $self->{volume}->get_zip_path(get_config('staging', 'zipfile')) . '.gpg';
+    return $self->{volume}->get_zip_path(get_config('staging', 'zipfile')) . "-$self->{name}.gpg";
```
aelkiss, line 95: This avoids collisions with encrypted zips left over from other storages. They *should* get cleaned up but don't always in practice.

lib/HTFeed/Stage/Collate.pm
```diff
@@ -125,4 +142,53 @@ sub clean_success {
...
+
+    my $mets_path = $volume->get_repository_mets_path;
+    die("METS missing (in $repo_path) after collate") unless $mets_path and -e $mets_path;
+
+    my $zipsize  = -s $zip_path;
+    my $zipdate  = $self->file_date($zip_path);
+    my $metssize = -s $mets_path;
+    my $metsdate = $self->file_date($mets_path);
+    my $sth      = get_dbh()->prepare($stmt);
+    get_logger()->trace("feed_audit: $zip_path / $zipdate / $zipsize bytes");
```
moseshll: Making sure you want a `get_logger` pattern here instead of the `$self->log_info` which seems to be preferred.
(May be able to get rid of the `use Log::Log4perl qw(get_logger);` if not needed)

moseshll (approved):
See inline suggestion re logging.

I'm pleased with all the cleanup happening here.

There still appears to be a brittle test but that's something that's happened sporadically to me for a long time, probably out of scope for this. I tend to suspect a race condition because it's a pretty simple test.
```
#   Failed test 'HTFeed::Storage::ObjectStore with encryption enabled stores the mets and encrypted zip'
#   at t/storage_object_store.t line 194.
```

lib/HTFeed/Stage/Collate.pm
aelkiss: I think I didn't implement a trace option but I could add it. It add some additional useful fields about the namespace/id/etc.

aelkiss:
re: brittle test I haven't seen it locally or in github. I can keep an eye out for it though.

lib/HTFeed/Stage/Collate.pm
aelkiss: Added log_trace and log_debug for future use.

aelkiss:
I will wait to merge & deploy; I'd like to get the schema updated in production and see about populating info using the existing audit stuff.

PR #186, opened 2026-06-16
ETT-316 issues with glacier expiration job

moseshll:
- Add `expire_versions.pl` invoked as worker
- Add `BackupExpirationBatch.pm` as core logic inside `expire_versions.pl`
  - No tests for this class, relying on the existing end-to-end tests to trickle down coverage
- `BackupExpiration`
  - `fork/exec` worker processes to handle deletion in parallel
  - Set default number of `BackupExpiration` workers to 8
  - Log some details on worker spawn and despawn
- Tests
  - Existing test suite left as intact as possible
  - Add test for unknown storage throwing exception
  - Add test with job size 1 so multiple workers are involved
  - Fix brittle tests with potential collisions from `old_random_timestamp` and `new_random_timestamp`

### DIFF
--- a/lib/HTFeed/BackupExpiration.pm
+++ b/lib/HTFeed/BackupExpiration.pm
@@ -44,6 +44,7 @@ sub new {
     custom_storage_config => 0,
     max_workers => 8,
     job_size => 10000,
+    limit => undef,
     @_
   };
 
@@ -83,7 +84,7 @@ sub run {
     close $fh;
   }
 
-  my $sth = get_dbh()->prepare($select_expired_sql);
+  my $sth = get_dbh()->prepare($self->select_expired_sql);
   my $versions_sth = get_dbh()->prepare($select_versions_sql);
 
   my $job = [];
@@ -175,6 +176,13 @@ sub spawn_worker {
     get_logger->trace("worker [$pid] started with $job_file (" . scalar(@$job) . " items)");
     $self->{workers}->{$pid} = $job_file;
   }
+}
+
+sub select_expired_sql {
+  my $self = shift;
+
+  my $limit_clause = (defined $self->{limit}) ? " LIMIT $self->{limit}" : '';
+  return $select_expired_sql . $limit_clause;
 }
 
 1;
--- a/lib/HTFeed/BackupExpiration.pm
+++ b/lib/HTFeed/BackupExpiration.pm
@@ -8,6 +8,7 @@ use HTFeed::Config qw(get_config);
 use HTFeed::DBTools qw(get_dbh);
 use HTFeed::Volume;
 
+use Carp ();
 use File::Spec ();
 use File::Temp ();
 use Log::Log4perl qw(get_logger);
@@ -41,12 +42,12 @@ sub new {
   my $self = {
     storage_name => undef,
     custom_storage_config => 0,
-    max_workers => 1,
+    max_workers => 8,
+    job_size => 10000,
     @_
   };
 
   unless ($self->{storage_name}) {
-    use Carp;
     Carp::croak "$class cannot be constructed without a storage name";
   }
 
@@ -67,8 +68,6 @@ sub new {
   bless($self, $class);
   return $self;
 }
-
-my $JOB_SIZE = 10000;
 
 sub run {
   my $self = shift;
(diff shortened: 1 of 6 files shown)


t/backup_expiration.t
```diff
@@ -116,9 +114,14 @@ describe "HTFeed::BackupExpiration" => sub {
   share my %vars;
   shared_examples_for "all storages" => sub {
     it "should create expiration object" => sub {
-      my $exp = HTFeed::BackupExpiration->new(storage_name => $vars{storage_name}, dry_run => 0);
+      my $exp = HTFeed::BackupExpiration->new(
+        storage_name => $vars{storage_name},
+        storage_config => $vars{storage_config},
```
moseshll, line 127: The only change here is to get storage config into the class under test, so it can pass it along (via YAML) to the worker processes that actually need it. Can't rely on patching the global config as we used to, when the work was being done in (or close to) the current process.

moseshll: The mere presence of `storage_config` passed to `new` is what prompts `BackupExpiration` to write the custom config to a file. I think it would be possible to just pass a flag that says in effect "this is custom config that won't be discoverable by a spawned process -- please write your config to a YAML file." But this is the mechanism that emerged, and it seems to work.

lib/HTFeed/BackupExpiration.pm
```diff
@@ -3,100 +3,177 @@ package HTFeed::BackupExpiration;
...
-    SELECT version
-    FROM feed_backups
-    WHERE deleted IS NULL
-      AND storage_name=?
-      AND namespace=?
-      AND id=?
-      AND version < DATE_FORMAT(DATE_SUB(NOW(), INTERVAL 180 DAY),"%Y%m%d%H%i%S")
-      ORDER BY version DESC
-SQL
+  # Write storage config to the temp directory for child processes to get at it.
```
aelkiss, line 76: I think I see why this is needed in terms of injecting config to the child process if we are doing `fork()` and `exec()` rather than just `fork()` and retaining the parents' config. It may be possible to set an env var to use a custom config (`HTFEED_CONFIG`) and write that out in the test? But if we need to do this I think it's OK.

lib/HTFeed/BackupExpiration.pm
```diff
@@ -3,100 +3,177 @@ package HTFeed::BackupExpiration;
...
-      $storage->{timestamp} = $version;
-      $storage->{zip_suffix} = '.gpg';
-      get_logger->trace("deleting archive for $volume->{namespace}.$volume->{objid} version $version" . $dry_run_text);
-      next if $dry_run;
-      unless ($storage->delete_objects) {
-        die "Unable to delete $volume->{namespace}.$volume->{objid}";
+    }
+  }
+  # Submit the leftovers if any
+  if (scalar @$job > 0) {
```
aelkiss, line 110: My initial reaction is that this seems like a lot of code for what we need to do here; while it looks correct at first glance it seems potentially brittle and reinventing the wheel.

Other options might be:
* writing chunks of ids to files and using xargs or parallel to handle each chunk with a configured max # of processes. Are there downsides in terms of this being too brittle?
* using argo workflow to parallelize after getting the chunks - probably don't need that level of scalability for this and it's still a fair amount of overhead
* using rabbitmq - also probably needs a fair amount of code although we have some of the machinery already. Workers could spawn & just stay there until the queue was empty. Could queue each deletion as its own job, although there's potential issues with very long queue lengths. If we needed to do this on an ongoing basis this could be preferable.
* migrating the code to ruby & using sidekiq. not ideal for a number of reasons - there's some logic in the storage classes that would need to be reimplemented - but again we could queue individual jobs and sidekiq is fairly well set up for this kind of thing.

In any case I guess the real question is if there benefits to the approach here that justify the amount of code?

aelkiss (commented):
Overall my sense is:

* writing out batches of IDs seems fine
* expire_versions.pl seems fine
* I think worth considering whether there are benefits to the fork/exec logic that justify the complexity here vs. other approaches

lib/HTFeed/BackupExpiration.pm
moseshll: @aelkiss My mantra has thus far been "put out the fire" so although both the Ruby and Argo (or both) directions are appealing, seems like that is something for later when it comes time for a major re-org. rabbitmq is something I have considered for a next iteration. I have relatively little hands-on experience with it but that can and should be remedied.

The benefit of what we have in this PR is that it should work _now_. Although I would probably want to add a `--limit` flag to the main executable (corresponding to a `LIMIT` SQL) to test a smallish tranche before committing to a months-long experiment.

aelkiss: Let's give this a try then. If something goes wrong here, the only real potential issue that I see is that we somehow manage to remove the backup version from storage without updating the database or vice versa, and that isn't really an issue of parallelization.

If we run into issues with this I'd be inclined to try the first option above (writing chunks of ids to files and using xargs or parallel to handle each chunk).

Using a `--limit` flag for the main executable seems like a good idea.

moseshll: `--limit` flag added

PR #188, opened 2026-06-29
ETT-316 issues with glacier expiration job

moseshll:
- `HTFeed::Storage::ObjectStore` calls `s3api delete-objects` with zip and mets instead of separate `rm` calls.
- Should halve the number of aws calls needed for Glacier expiration.
- If successful should provide a model for deleting up to 1k items at a time.

### DIFF
--- a/lib/HTFeed/Storage/ObjectStore.pm
+++ b/lib/HTFeed/Storage/ObjectStore.pm
@@ -11,6 +11,7 @@ use HTFeed::StorageAudit::ObjectStore;
 use Log::Log4perl qw(get_logger);
 use MIME::Base64 qw(decode_base64);
 use POSIX qw(strftime);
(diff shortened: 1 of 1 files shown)


aelkiss (approved):
Let's go for it.

One option to try it without needing to deploy (probably worthwhile, especially since we will likely want to increase the number of objects per call):
* Start a pod with the necessary secrets etc with the command `sleep infinity` (e.g. by changing the command for the cron job and manually triggering the job)
* Copy the updated script into /tmp on the pod (with `kubectl cp`)
* Exec into the pod
* Run the script using nohup

aelkiss:
I'm going to go ahead and merge and deploy, and we can try this, although I suspect we'll probably want to still increase the number of objects per API call.

PR #189, opened 2026-06-30
Log only errors and "collate succeeds" messages to feed_log

aelkiss:
This should cut down significantly on the noise in feed_log, although we still ought to rethink how reporting that uses feed_log works.


moseshll (approved):
Works here after resolution of minor kinks.

PR #190, opened 2026-07-06
ETT-316 issues with glacier expiration job (`mass_delete` edition)

moseshll:
- Add a `mass_delete` class method to the `ObjectStore` and `PrefixedVersions` storage classes.
   - Intended to replace the `delete_objects` instance method.
 - Add companion `object_keys` instance method so these can be amassed at a higher level and removed via `mass_delete`.
 - `BackupExpirationBatch` can still take an arbitrarily large batch size, but due to s3api delete-objects limitation of 1k objects it iterates sub-batches of just under 1k. This way there's no need to mess with the job-size parameter in the parent binary.
   - Switch to a single `mass_delete` storage call per subbatch.

### DIFF
--- a/lib/HTFeed/BackupExpiration.pm
+++ b/lib/HTFeed/BackupExpiration.pm
@@ -127,19 +127,16 @@ sub run {
 sub wait_for_available_worker {
   my $self = shift;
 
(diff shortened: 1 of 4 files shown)


aelkiss (approved):
Let's give it a try!

